PR #5, opened 2022-01-21
Add Entities.include

imjn:
Hello :)

## What

While implementing this plugin into our project, I found that `Entities.skip` is very hard to maintain.
Therefore, I'd like to implement `include` which works completely opposite way from `skip`.

## Background

We are currently just replacing hard-coded entities with CreateAPI generated entities, but of course, it is difficult to replace hundreds of entities at once. Therefore, we plan to replace them gradually. In this context, `include` is much more useful than `skip`.

## Changes
- Renamed `skip` to `exclude` to provide a kind of consistency with `include`. If you have any other suggestions, please :)
- Added `include`
- Throw error when both of `include` and `exclude` have value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,8 +80,11 @@ isSwiftLintDisabled: true
 fileHeader: null
 
 entities:
-  # Skipped entities, e.g. ["SimpleUser"]
-  skip: []
+  # Excluded entities, e.g. ["SimpleUser"]
+  # `exclude` and `include` can't be used together
+  exclude: []
+  # Included entities, e.g. ["SimpleUser"]
+  include: []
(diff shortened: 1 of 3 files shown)


kean:
Looks great, thanks for the contribution.

I also added a ticket to replace "skip" everywhere https://github.com/kean/CreateAPI/issues/6 as a reminder.

kean:
> I also added a ticket to replace "skip" everywhere https://github.com/kean/CreateAPI/issues/6 as a reminder.

Fixed in [0.0.2](https://github.com/kean/CreateAPI/releases/tag/0.0.2)

PR #10, opened 2022-01-27
Discriminator Support

PhilipTrauner:
Adds initial support for decoding `oneOf` entities with a [discriminator](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) attribute.

Using the existing property type name machinery when constructing the discriminator is still outstanding. 
Would really appreciate some help with that, because I struggled to find a clean way to call into one of the name canonicalization functions.

<details>
	<summary>Generated Code</summary>

```swift
public struct A: Codable {
    public var kind: String

    public init(kind: String) {
        self.kind = kind
    }
}

public struct B: Codable {
    public var kind: String

    public init(kind: String) {
        self.kind = kind
    }
}

public struct C: Codable {
    public var kind: String

    public init(kind: String) {
        self.kind = kind
    }
}

public struct Container: Codable {
    public var content: Content

    public enum Content: Codable {
        case a(A)
        case b(B)
        case c(C)

            struct Discriminator: Decodable {
                let kind: String
            }

            let container = try decoder.singleValueContainer()

            switch try? container.decode(Discriminator.self) {
            case .some(let inner) where inner.kind == "a":
                self = .a(try container.decode(A.self))
            case .some(let inner) where inner.kind == "b":
                self = .b(try container.decode(B.self))
            case .some(let inner) where inner.kind == "c":
                self = .c(try container.decode(C.self))
            default:
                throw DecodingError.dataCorruptedError(in: container, debugDescription: "Failed to initialize `oneOf`")
            }
        }

            var container = encoder.singleValueContainer()
            switch self {
            case .a(let value): try container.encode(value)
            case .b(let value): try container.encode(value)
            case .c(let value): try container.encode(value)
            }
        }
    }

    public init(content: Content) {
        self.content = content
    }
}

struct StringCodingKey: CodingKey, ExpressibleByStringLiteral {
    private let string: String
    private var int: Int?

    var stringValue: String { return string }

    init(string: String) {
        self.string = string
    }

    init?(stringValue: String) {
        self.string = stringValue
    }

    var intValue: Int? { return int }

    init?(intValue: Int) {
        self.string = String(describing: intValue)
        self.int = intValue
    }

    init(stringLiteral value: String) {
        self.string = value
    }
}
```
</details>

<details>
	<summary>Schema</summary>

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Discriminator

paths:
  /container:
    get:
      responses:
        "200":
          description: ""
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Container"
components:
  schemas:
    A:
      type: object
      required:
        - kind
      properties:
        kind:
          type: string
    B:
      type: object
      required:
        - kind
      properties:
        kind:
          type: string
    C:
      type: object
      required:
        - kind
      properties:
        kind:
          type: string

    Container:
      type: object
      required:
        - content
      properties:
        content:
          oneOf:
            - $ref: "#/components/schemas/A"
            - $ref: "#/components/schemas/B"
            - $ref: "#/components/schemas/C"
          discriminator:
            propertyName: kind
            mapping:
              a: "#/components/schemas/A"
              b: "#/components/schemas/B"
              c: "#/components/schemas/C"
```
</details>

### DIFF
--- a/Sources/CreateAPI/Generator/Generator+Schemas.swift
+++ b/Sources/CreateAPI/Generator/Generator+Schemas.swift
@@ -409,10 +409,40 @@ extension Generator {
         }
     }
 
+    private func makeDiscriminator(info: JSONSchemaContext, context: Context) -> Discriminator? {
+        if let discriminator = info.discriminator {
+            var mapping: [String: TypeIdentifier] = [:]
+
+            if let innerMapping = discriminator.mapping {
+                
+                // TODO: Use type name machinery instead
+                mapping = innerMapping.mapValues { value in 
+                    return .userDefined(
+                        name: makeTypeName(String(value.dropFirst("#/components/schemas/".count)))
+                    )
+                }
+            }
+
+            return .init(
+                propertyName: discriminator.propertyName,
+                mapping: mapping
+            )
+        }
+
+        return .none
+    }
+
     // MARK: - oneOf/anyOf/allOf
     
     private func makeEntity(name: TypeName, type: EntityType, info: JSONSchemaContext, context: Context) -> (EntityDeclaration, Context) {
-        let entity = EntityDeclaration(name: name, type: type, metadata: DeclarationMetadata(info), isForm: context.isFormEncoding, parent: context.parents.last)
+        let entity = EntityDeclaration(
+            name: name, 
+            type: type, 
+            metadata: DeclarationMetadata(info), 
+            isForm: context.isFormEncoding,
+            discriminator: makeDiscriminator(info: info, context: context),
+            parent: context.parents.last
+        )
         let context = context.map { $0.parents.append(entity) }
         return (entity, context)
     }
--- a/Sources/CreateAPI/Generator/Generator+Schemas.swift
+++ b/Sources/CreateAPI/Generator/Generator+Schemas.swift
@@ -409,17 +409,22 @@ extension Generator {
         }
     }
 
-    private func makeDiscriminator(info: JSONSchemaContext, context: Context) -> Discriminator? {
+    private func makeDiscriminator(info: JSONSchemaContext, context: Context) throws -> Discriminator? {
         if let discriminator = info.discriminator {
             var mapping: [String: TypeIdentifier] = [:]
 
             if let innerMapping = discriminator.mapping {
-                
-                // TODO: Use type name machinery instead
-                mapping = innerMapping.mapValues { value in 
-                    return .userDefined(
-                        name: makeTypeName(String(value.dropFirst("#/components/schemas/".count)))
-                    )
+                for (key, value) in innerMapping {
+                    let stripped = String(value.dropFirst("#/components/schemas/".count))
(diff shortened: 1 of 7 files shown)


Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -406,10 +406,40 @@ extension Generator {
...
     }
 
+    private func makeDiscriminator(info: JSONSchemaContext, context: Context) -> Discriminator? {
+        if let discriminator = info.discriminator {
+            var mapping: [String: TypeIdentifier] = [:]
+
+            if let innerMapping = discriminator.mapping {
+                
+                // TODO: Use type name machinery instead
+                mapping = innerMapping.mapValues { value in
```
kean, line 416: I think this should work:

```swift
mapping = innerMapping.mapValues { value in
    guard let key = OpenAPI.ComponentKey(rawValue: value) else {
        throw GeneratorError("TBD") // This should never happen
    }
    // This takes care of options.rename.
    // It _doesn't_ take options.entities.include / exclude into the account.
    // Ideally CreateAPI should throw an error (or warning if --strict is disabled) if
    // the type is exlucded (neat-picking).
    guard let name = getTypeName(for: key) else {
        // TODO:
    }
    return .userDefined(name: name)
}
```

kean:
I suggest to cleanup the generated code a bit. I think this should be functionally equivalent: 

```swift
switch try container.decode(Discriminator.self) {
case "a": self = .a(try container.decode(A.self))
// ...
}
```

Sources/CreateAPI/Generator/Generator+Schemas.swift
PhilipTrauner:
```swift
guard let key = OpenAPI.ComponentKey(rawValue: value) else {
    throw GeneratorError("TBD") // This should never happen
}
```
 
The `rawValue:` constructor enforces that the provided value matches this regex: `^[a-zA-Z0-9\.\-_]+$`.
At this stage the mapping values are still prefixed with `"#/components/schemas/"`, which does not conform to said regex.

kean: You are right. I just tested it and yes, the string value should be the component name (without "#/components/schemas/").

```swift
String(value.dropFirst("#/components/schemas/".count))
```

That's a nice approach for removing the prefix. I would've usually used `replacingOccurrences`, but this is much better.

kean:
I fixed the test target in `main`. I need to restructure the package a bit. I'll do it once this PR is merged so you don't have to move anything / resolve conflicts.

kean:
Thanks, @PhilipTrauner. Merging it.

I'll address the following later today and add the expected test output to the repo:

```swift
// TODO: Find it automatically
var projectPath = ("~/Developer/CreateAPI/" as NSString).expandingTildeInPath
```

It's needed for "snapshot" testing. The idea is that when you add a new test, and it can't find the expected files in the repo, it generates them for you to commit.

PR #14, opened 2022-02-01
Add entities name template

imjn:
# WHY

Related to https://github.com/kean/CreateAPI/issues/3

We already have many entities with the same names as the OpenAPI spec.
Of course, you can apply namespace to use generated entities and existing entities together. But we wanted to make it easier to distinguish.

For example

`Book` -> `BookDTO`

# WHAT

Added `namePlaceholder` for entities. If you set a value like `hoge*Hoge`, the entities will be like `hogeEntityHoge`.

If there's better practice, please give me some advice.

Cheers.


kean:
Hey @imjn. Yeah, I like this approach.

I'm using something similar with filenames:

```
  --filename-template <filename-template>
                          Example: "%0.generated.swift" will produce files with the following names:
                          "Paths.generated.swift". (default: %0.swift)
```

I suggest to name the parameter `entityNameTemplate` and using `%0` as a parameter for consistency.

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -100,11 +100,25 @@ extension Generator {
     }
     
     private func shouldGenerate(name: String) -> Bool {
-        if !options.entities.include.isEmpty {
-            return options.paths.include.contains(name)
+        var entitiesToInclude: Set<String> {
+            if let placeholder = options.entities.namePlaceholder {
+                return Set(options.entities.include.map { placeholder.replacingOccurrences(of: "*", with: $0) })
```
kean, line 105: Please remove duplication. I think a good place for name generation is the existing `getTypeName` method.

Sources/CreateAPI/Generate.swift
```diff
@@ -142,6 +146,11 @@ struct Generate: ParsableCommand {
         return GenerateOptions(options)
     }
     
+    private func applyTemplateToEntityIncludeAndExclude(options: GenerateOptions) {
```
imjn, line 149: I think this is not smart but needed when you use entities.include or entities.exclude. Is there any other good solutions for this?

kean:
Thanks for addressing the comments! I'm going to merge it and maybe tweak it a little bit in `develop`.

PR #15, opened 2022-02-03
Strip parent name of enum cases within nested objects

PhilipTrauner:
Strips the parent name of enum cases within objects that are `oneOf` / `allOf` / `anyOf` of nested references.

Given:

```
Parent
├── ParentA
└── ParentB
```

Disabled:

```swift
public enum Parent: Codable {
    case parentA(ParentA)
    case parentB(ParentB)
}
``` 

Enabled:

```swift
public enum Parent: Codable {
    case a(ParentA)
    case b(ParentB)
}
``` 

This is primarily useful for union types.


kean:
That's brilliant. That's exactly the kind of features I think CreateAPI should have.

PR #19, opened 2022-02-07
Added imports option for entities

JanC:
As mentioned in https://github.com/kean/CreateAPI/issues/18#issuecomment-1031208978, this PR adds the `imports` option to the Entities.

**Why?**
Similar to the `paths.imports`,  this pretty much allows the user the import any modules in the generated entities files


kean:
Great, thanks!

PR #25, opened 2022-02-17
Fix `String` type with `byte` format

mattia:
When the spec defines a string property with a byte format at the moment a `String` property is generated, but a `Data` property should  be generated instead.

With input:
```yaml
test_property:
  type: string
  format: byte
```

Before:
```swift
var testProperty: String
```

After:
```swift
var testProperty: Data
```


kean:
Thanks for making the change! I appreciate it.

PR #27, opened 2022-03-10
Fix allOf decoding issue

imjn:
## WHAT

When we have this kind OpenAPI schema ↓
```JSON
"Dog": {
  "allOf": [
    {
      "type": "object",
      "additionalProperties": false,
      "required": ["type"],
      "properties": {
        "type": {
          "type": "string",
          "enum": ["dog"]
        }
      }
    },
    {
      "$ref": "#/components/schemas/BaseDog"
    }
  ]
}
```
it fails to decode `type` because currently, CreateAPI generates ↓
``` Swift
    self.type = try `Type`(from: decoder) # ← THIS IS WRONG
    self.baseDog = try BaseDog(from: decoder)
}
```

It should be
```Swift
    let values = try decoder.container(keyedBy: StringCodingKey.self)
    self.type = try values.decode(`Type`.self, forKey: "type") # ← THIS IS CORRECT
    self.baseDog = try BaseDog(from: decoder)
}
```

In this PR, I've added `isInlineed` to `Property` in order to differentiate `Type` and `BaseDog` generated codes in the example above.

Also, when an object in `allOf` has a reference type property, it fails because currently it is generated as

```yml
# spec
Dog:
    allOf:
      - $ref: "#/components/schemas/Animal"
      - type: object
        properties:
          breed:
            type: string
          image:
            $ref: "#/components/schemas/Image"
```
```Swift
public struct Dog: Codable {
    public var animal: Animal
    public var breed: String?
    public var image: Image?

    public init(animal: Animal, breed: String? = nil, image: Image? = nil) {
        self.animal = animal
        self.breed = breed
        self.image = image
    }

        let values = try decoder.container(keyedBy: StringCodingKey.self)
        self.animal = try Animal(from: decoder)
        self.breed = try values.decodeIfPresent(String.self, forKey: "breed")
        self.image = try Image(from: decoder) # ← THIS IS WRONG
    }

        var values = encoder.container(keyedBy: StringCodingKey.self)
        try values.encode(animal, forKey: "animal")
        try values.encodeIfPresent(breed, forKey: "breed")
        try values.encodeIfPresent(image, forKey: "image")
    }
}
```
but should be
```Swift
public struct Dog: Codable {
    public var animal: Animal
    public var breed: String?
    public var image: Image?

    public init(animal: Animal, breed: String? = nil, image: Image? = nil) {
        self.animal = animal
        self.breed = breed
        self.image = image
    }

        let values = try decoder.container(keyedBy: StringCodingKey.self)
        self.animal = try Animal(from: decoder)
        self.breed = try values.decodeIfPresent(String.self, forKey: "breed")
        self.image = try values.decodeIfPresent(Image.self, forKey: "image") # ← THIS IS CORRECT
    }

        var values = encoder.container(keyedBy: StringCodingKey.self)
        try values.encode(animal, forKey: "animal")
        try values.encodeIfPresent(breed, forKey: "breed")
        try values.encodeIfPresent(image, forKey: "image")
    }
}
```

## HOW

- [x] Added some tests
- [x] Added `isInlined` to `Property`.

### DIFF
--- a/Sources/CreateAPI/Generator/Declarations.swift
+++ b/Sources/CreateAPI/Generator/Declarations.swift
@@ -159,7 +159,7 @@ struct Property {
     // A nested declaration required used as a property type
     var nested: Declaration?
     // If the schema is inlined by `allOf`. This is currently used only for the generation of decoders.
-    var isInlined: Bool
+    var isInlined: Bool?
 }
 
 protocol Declaration {
--- a/Sources/CreateAPI/Generator/Declarations.swift
+++ b/Sources/CreateAPI/Generator/Declarations.swift
@@ -90,8 +90,7 @@ indirect enum TypeIdentifier: CustomStringConvertible, Hashable {
         switch self {
         case .array(let element): return "[\(element)]"
         case .dictionary(let key, let value): return "[\(key): \(value)]"
-        case .userDefined(let name): return name.rawValue
-        case .builtin(let name): return name.rawValue
+        case .userDefined(let name), .builtin(let name): return name.rawValue
         }
     }
 }
@@ -159,6 +158,8 @@ struct Property {
     var metadata: DeclarationMetadata?
     // A nested declaration required used as a property type
     var nested: Declaration?
+    // If the schema is inlined by `allOf`. This is currently used only for the generation of decoders.
+    var isInlined: Bool
 }
 
 protocol Declaration {
--- a/Sources/CreateAPI/Generator/Generator+Render.swift
+++ b/Sources/CreateAPI/Generator/Generator+Render.swift
@@ -80,13 +80,8 @@ extension Generator {
             case .allOf:
                 var needsValues = false
                 let decoderContents = properties.map {
-                    if case .userDefined = $0.type {
-                        if $0.isInlined ?? false {
-                            needsValues = true
-                            return templates.decode(property: $0, isUsingCodingKeys: false)
-                        } else {
-                            return templates.decodeFromDecoder(property: $0)
-                        }
+                    if case .userDefined = $0.type, $0.isInlined != true {
+                        return templates.decodeFromDecoder(property: $0)
                     } else {
                         needsValues = true
                         return templates.decode(property: $0, isUsingCodingKeys: false)
--- a/Sources/CreateAPI/Generator/Generator+Render.swift
+++ b/Sources/CreateAPI/Generator/Generator+Render.swift
@@ -81,7 +81,7 @@ extension Generator {
                 var needsValues = false
                 let decoderContents = properties.map {
                     if case .userDefined = $0.type {
-                        if $0.isInlined {
+                        if $0.isInlined ?? false {
                             needsValues = true
                             return templates.decode(property: $0, isUsingCodingKeys: false)
                         } else {
--- a/Sources/CreateAPI/Generator/Generator+Render.swift
+++ b/Sources/CreateAPI/Generator/Generator+Render.swift
@@ -81,7 +81,12 @@ extension Generator {
                 var needsValues = false
                 let decoderContents = properties.map {
                     if case .userDefined = $0.type {
-                        return templates.decodeFromDecoder(property: $0)
+                        if $0.isInlined {
+                            needsValues = true
+                            return templates.decode(property: $0, isUsingCodingKeys: false)
+                        } else {
+                            return templates.decodeFromDecoder(property: $0)
+                        }
                     } else {
                         needsValues = true
                         return templates.decode(property: $0, isUsingCodingKeys: false)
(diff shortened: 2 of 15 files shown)


kean:
Thanks for working on this. I merged the other two MRs and also fixed the remaining tests in the main branch.

Tests/CreateAPITests/Specs/edgecases.yaml
```diff
@@ -1001,6 +1003,15 @@ components:
...
           default: red
+    Image:
+      type: object
+      required:
+        - id
+        - url
+      id:
+        - type: string
+      url:
+        - type: string
```
liamnichols, line 1014: I think the yaml formatting is a little off here which is why the `type` is being ignored for the generated `Image` (it's `AnyJSON` when it should be `String`)

iirc, the `-` denotes an array item, so instead we want this:

```suggestion
    Image:
      type: object
      required:
        - id
        - url
      id:
        type: string
      url:
        type: string
```

Sources/CreateAPI/Generator/Generator+Render.swift
```diff
@@ -81,7 +81,12 @@ extension Generator {
                 var needsValues = false
                 let decoderContents = properties.map {
                     if case .userDefined = $0.type {
-                        return templates.decodeFromDecoder(property: $0)
+                        if $0.isInlined ?? false {
+                            needsValues = true
+                            return templates.decode(property: $0, isUsingCodingKeys: false)
```
liamnichols, line 86: I think we can avoid the nested if (and the duplication on the lines below) by flipping this check and including it on the outer if statement.

Would this produce the same result?

```swift
if case .userDefined = $0.type, $0.isInlined != true {
    return templates.decodeFromDecoder(property: $0)
} else {
    needsValues = true
    return templates.decode(property: $0, isUsingCodingKeys: false)
}
```

Sources/CreateAPI/Generator/Declarations.swift
```diff
@@ -159,6 +158,8 @@ struct Property {
     var metadata: DeclarationMetadata?
     // A nested declaration required used as a property type
     var nested: Declaration?
+    // If the schema is inlined by `allOf`. This is currently used only for the generation of decoders.
+    var isInlined: Bool?
```
liamnichols, line 162: Quick question, what is the difference in meaning of this property between when the value is `nil` and when it's `false`?

liamnichols (commented):
This looks great @imjn 💪 

Did you also want to replicate the first issue that you described in the PR description within the specs/tests? Or do you think just adding `Image` to `Animal` is sufficient like you have done already? I'm just trying to understand if the current tests are covering all of the behavioural changes 🙏

Sources/CreateAPI/Generator/Declarations.swift
imjn: I made this `isInlined` optional because this is only used for generating schemas but the struct `Property` is also used for the generation of Paths, therefore I thought setting `false` for this even though we don't know if it's actually inlined or not is kinda misleading. I defined `true` or `false` during the generation of the schemas because it needs to know if it is in `allOf` generation flow or not.

So the answer to the question would be `false` means it's not inlined and `nil` means we don't know if it's inlined or not and don't need to know either.

Sources/CreateAPI/Generator/Generator+Render.swift
imjn: Thanks! Fixed in [b276bb8](https://github.com/CreateAPI/CreateAPI/pull/27/commits/b276bb8d8c2b3204da8b43ac1890dcf66bab7259).

imjn:
Added test coverage for the first issue that I described in the description [6ebb9f5](https://github.com/CreateAPI/CreateAPI/pull/27/commits/6ebb9f591032054799e5f9c2458ad02097dc4aef).
This is for `Enum` decoding issue in `allOf`.

Tests/CreateAPITests/Expected/edgecases-yaml-config/Sources/Entities.swift
```diff
@@ -281,6 +291,21 @@ public struct Animal: Codable {
     }
 }
 
+public struct Image: Codable {
+    public var identifier: AnyJSON
+    public var url: AnyJSON
```
liamnichols, line 296: Do you need to do anything to re-generate this? Looking at the schema, I think these should be `String` and not `AnyJSON` but I can't figure out why 🤔

Tests/CreateAPITests/Specs/edgecases.yaml
```diff
@@ -1001,6 +1007,15 @@ components:
...
           default: red
+    Image:
+      type: object
+      required:
+        - id
+        - url
+      id:
+        type: string
+      url:
+        type: string
```
liamnichols, line 1018: Ah, I figured it out... They need to be nested under `properties`

```suggestion
      properties:
        id:
          type: string
        url:
          type: string
```

liamnichols (commented):
Looking great @imjn! Thanks again for solving this 💪 Just one tweak to the schema and then I think its good to go 🚀

liamnichols (approved):
Great! Lets merge this 🚀

PR #31, opened 2022-03-27
Fixed optional decoding from decoder

imjn:
`try Entity(from: decoder)` should be `try? Entity(from: decoder)` if the property is optional?


liamnichols (commented):
Thanks @imjn! 

Do you think that we can get a test in place for this too? Perhaps we can extend one of the existing schemas to add an extra component that replicates the scenario?

imjn:
Sorry, I found this change is not needed.

PR #36, opened 2022-05-13
Support multiple discriminator mappings to share one type

imjn:
## Background

Currently, CreateAPI only supports a unique type for each mapping case.
For example, if we have

```yml
Container:
      type: object
      required:
        - content
      properties:
        content:
          oneOf:
            - $ref: "#/components/schemas/A"
            - $ref: "#/components/schemas/B"
            - $ref: "#/components/schemas/C"
          discriminator:
            propertyName: kind
            mapping:
              a: "#/components/schemas/A"
              b: "#/components/schemas/B"
              c: "#/components/schemas/C"
              d: "#/components/schemas/A"
```

it only generates `.a` but `.d` is not generated.
This is because we only use the first case with the property type.

https://github.com/CreateAPI/CreateAPI/blob/a59c76632f3199ddb016447dc04ceb83102d3f61/Sources/CreateAPI/Generator/Templates.swift#L314

## How

In this PR, I have fixed the issue above by checking the `mapping` keys and updated the discriminator test case for the case when two enum cases share one type.

Hope the changes make sense. Let me know if you have better options 🙇🏽

### DIFF
--- a/Sources/CreateAPI/Generator/Generator+Render.swift
+++ b/Sources/CreateAPI/Generator/Generator+Render.swift
@@ -41,7 +41,20 @@ extension Generator {
                 contents.append(templates.initializer(properties: properties))
             }
         case .oneOf:
-            contents.append(properties.map(templates.case).joined(separator: "\n"))
+            if let discriminator = decl.discriminator {
+                var statements: [String] = []
+                for property in properties {
+                    let correspondingMappings = discriminator.mapping.filter { $1 == property.type }.sorted { $0.key < $1.key }
+                    if !correspondingMappings.isEmpty {
+                        for mapping in correspondingMappings {
+                            statements.append("case \(mapping.key)(\(property.type))")
+                        }
+                    }
+                }
+                contents.append(statements.joined(separator: "\n"))
+            } else {
+                contents.append(properties.map(templates.case).joined(separator: "\n"))
+            }
             contents += decl.nested.map(render)
         }
 
@@ -102,11 +115,11 @@ extension Generator {
                     }
                 }
                 if decl.protocols.isEncodable {
-                    contents.append(templates.encodeOneOf(properties: properties))
+                    contents.append(templates.encodeOneOf(properties: properties, discriminator: decl.discriminator))
                 }
             }
         }
-        
+
         // TODO: Refactor
         var protocols = decl.protocols
         if decl.isForm {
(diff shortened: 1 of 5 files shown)


Sources/CreateAPI/Generator/Templates.swift
```diff
@@ -343,10 +346,23 @@ final class Templates {
...
     }    
     
-    func encodeOneOf(properties: [Property]) -> String {
-        let statements = properties.map {
-            "case .\($0.name)(let value): try container.encode(value)"
+    func encodeOneOf(properties: [Property], discriminator: Discriminator?) -> String {
+        var statements: [String] = []
+        if let discriminator = discriminator {
+            for property in properties {
+                let correspondingMappings = discriminator.mapping.filter { $1 == property.type }.sorted { $0.key < $1.key }
```
liamnichols, line 353: I see that this has been repeated three times so do you think we can extract it out to a computed property somewhere? Maybe on `Discriminator`?

Sources/CreateAPI/Generator/Templates.swift
```diff
@@ -343,10 +346,23 @@ final class Templates {
...
+                if !correspondingMappings.isEmpty {
+                    for mapping in correspondingMappings {
+                        statements.append("case .\(mapping.key)(let value): try container.encode(value)")
+                    }
+                }
```
liamnichols, line 358: Passing an empty collection to a `for` loop isn't an issue so we can avoid the `!isEmpty` check here

```suggestion
                for mapping in correspondingMappings {
                    statements.append("case .\(mapping.key)(let value): try container.encode(value)")
                }
```

Sources/CreateAPI/Generator/Generator+Render.swift
```diff
@@ -41,7 +41,20 @@ extension Generator {
...
-            contents.append(properties.map(templates.case).joined(separator: "\n"))
+            if let discriminator = decl.discriminator {
+                var statements: [String] = []
+                for property in properties {
+                    let correspondingMappings = discriminator.mapping.filter { $1 == property.type }.sorted { $0.key < $1.key }
+                    if !correspondingMappings.isEmpty {
+                        for mapping in correspondingMappings {
+                            statements.append("case \(mapping.key)(\(property.type))")
+                        }
+                    }
```
liamnichols, line 52:
```suggestion
                    for mapping in correspondingMappings {
                        statements.append("case \(mapping.key)(\(property.type))")
                    }
```

liamnichols (commented):
Thanks for picking this up! It's looking good 💪💪💪

Sources/CreateAPI/Generator/Templates.swift
imjn: Thanks for your reviews! I couldn't find a way to make it with a computed property so I went with a function. Is this alright?
[446ea60](https://github.com/CreateAPI/CreateAPI/pull/36/commits/446ea605b9304ba50c7397a8ad97239f38ea44c8)

liamnichols (approved):
Nice one @imjn, great work! 🚀

PR #39, opened 2022-06-10
Treat single element allOf/oneOf/anyOf schemas as the nested schema

liamnichols:
# Background

In our schema, we frequently do something like the following:

```json
{
  "type": "object",
  "required": [
    "user"
  ],
  "properties": {
    "user": {
      "nullable": false,
      "allOf": [
        {
          "$ref": "#/components/schemas/User"
        }
      ]
    }
  }
}
```

In the above example, the `user` property needs to be nullable, but the schema of the property is actually a reference to a non-nullable component. The semi-official (but not ideal) way that was established to handle this case is to nest the `$ref` in an `allOf`/`anyOf`/`oneOf` array and mark it as `nullable` (you can't have other annotations alongside `$ref`).

While this approach works in our Android code generator, and our server-side validator, it produces unexpected results in CreateAPI when using `isInliningPropertiesFromReferencedSchemas` because CreateAPI will attempt to inline the `allOf` as a new type instead of referencing the actual type as an optional property.

# Changes

To reproduce the issue locally, I first needed to setup some test coverage since we hadn't tested it before now. To do that, I've added a new inlining.yml schema and setup a few scenarios that test this behaviour in general.

Finally, I applied the actual fix in 48237df


Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -193,6 +193,10 @@ extension Generator {
             return try makeObject(name: name, info: info, details: details, context: context)
         case .array(let info, let details):
             return try makeTypealiasArray(name: name, info: info, details: details, context: context)
+        case .all(let schemas, _) where schemas.count == 1,
```
imjn, line 196: Adding some comment that describes why we do like this might help others in the future...?

imjn (approved):
The changes and the tests both look great to me 👍🏽 Thank you 🥇

PR #41, opened 2022-06-10
Revert "Fix `String` type with `byte` format"

### LINKED ISSUE #29: Is `Data` the correct type for properties with the `format: byte` modifier?
The type of properties with the `format: byte` modifier has been changed to `Data` in PR #25.

However, according to the Swagger documentation, they are defined as Base64-encoded characters.
So it would be correct to treat it as a `String`, wouldn't it?

https://swagger.io/docs/specification/data-models/data-types/#format
> An optional format modifier serves as a hint at the contents and format of the string. OpenAPI defines the following built-in string formats:
...
byte – base64-encoded characters, for example, U3dhZ2dlciByb2Nrcw==

liamnichols:
Reverts #25
Closes #29

> According to the Swagger documentation, they are defined as Base64-encoded characters.
> So it would be correct to treat it as a `String`, wouldn't it?
> 
> 
> https://swagger.io/docs/specification/data-models/data-types/#format
> > An optional format modifier serves as a hint at the contents and format of the string. OpenAPI defines the following built-in string formats:
> ...
> byte – base64-encoded characters, for example, U3dhZ2dlciByb2Nrcw==

cc @simorgh3196, @mattia

### DIFF
--- a/Sources/CreateAPI/Generator/Generator+Schemas.swift
+++ b/Sources/CreateAPI/Generator/Generator+Schemas.swift
@@ -242,8 +242,6 @@ extension Generator {
             return .builtin("URL")
         case .other(let other) where other == "uuid":
             return .builtin("UUID")
-        case .byte:
-          return .builtin("Data")
         default: break
         }
         return .builtin("String")


liamnichols:
Actually, I don't think that this needs reverting. I'll share some notes in the issue.

PR #43, opened 2022-06-27
Support Linux

liamnichols:
# Background

While most use-cases evolve around Apple Platforms, there are still benefits to being able to use CreateAPI to generate Swift source code on Linux machines.

For example, a workflow could produce and publish the Swift package whenever a shared OpenAPI schema document is updated, but it might be more cost effective to run that flow on Linux instead of macOS machines.

# Changes

Most of CreateAPI relies entirely on Swift itself and nothing within the Apple ecosystem so this was a pretty straightforward change. 

The only limitation here is that the FileWatcher dependency that is used for the `--watch` command does not support Linux. As a result, I decided to conditionally exclude this functionality when building for Linux.

While ideally I would like to support Swift 5.5 on Linux too, it for some reason is not possible to access `Unicode.Scalar.Properties.isEmoji` which causes the build to fail. I didn't find much on the internet besides this: https://forums.swift.org/t/swift-include-icu4c/39790/12. My focus here is mostly on maintaining a minimum level of Linux support rather than forcing 100% compatibility in unreasonable ways so I think its fine to skip 5.5

### DIFF
--- a/Sources/CreateAPI/Generate.swift
+++ b/Sources/CreateAPI/Generate.swift
@@ -9,7 +9,6 @@ import Yams
 
 // TODO: Generate `README.md` for package (see `info`)
 // TODO: Disable sandbox
-// TODO: Add Linux support
 // TODO: Add OpenAPI 3.1 support
 
 struct Generate: ParsableCommand {


.github/workflows/ci.yml
```diff
@@ -30,3 +30,13 @@ jobs:
         run: swift test
       - name: Compile Generated Code
         run: ./Scripts/test-generated.sh
+  linux:
+    name: Linux
+    runs-on: ubuntu-latest
+    container: swift:5.6
```
liamnichols, line 36: Maybe I should run this on Swift 5.5 as well? Same as we are doing with the Xcode matrix?

PR #45, opened 2022-06-29
Removed TypealiasDeclaration return for allOf making

imjn:
## WHY

CreateAPI is not generating cases for `oneOf` properly which has only one property in its `allOf`.

Example.
```JSON
{
  "BlogPostType": {
    "oneOf": [
      {
        "$ref": "../schema.json#/components/schemas/BlogPost"
      }
      ...
    ]
  }
}

{
  "BlogPost": {
    "allOf": [
      {
        "$ref": "../global_web.json#/components/schemas/BlogPostItem"
      },
      {
        "type": "object"
      }
    ]
  }
}
```
In this case, the `case .blogPost` will not be generated but `case .blogPostType` will be as it uses the `TypealiasDeclaration` with the name of the parent property.

## HOW

Removed the `properties.count` checking and returning `TypealiasDeclaration` which doesn't make sense in `makeAllOf` function 🤔 

## TESTS

TBA


Tests/Support/Snapshots/discriminator/Sources/Entities/ContainerWithDirectOneOf.swift
```diff
@@ -0,0 +1,42 @@
+// Generated by Create API
+// https://github.com/CreateAPI/CreateAPI
+
+import Foundation
+
+public enum ContainerWithDirectOneOf: Codable {
+    case a(A)
+    case three(Three)
+    case four(Four)
```
imjn, line 9: This `case four(Four)` becomes `case string(String)` without the change in this PR

imjn:
Needed changes will be included in https://github.com/CreateAPI/CreateAPI/pull/143

PR #46, opened 2022-06-29
Fix comment generation when using other kind of linebreaks

### LINKED ISSUE #44: Support escape character \r in comments
Hello, first of all, thank you for providing a awesome tool.

Recently, I tried to generate OpenAPI provided by [Jellyfin](https://github.com/jellyfin/jellyfin) with CreateAPI, but some comments had problems.
Please refer to the screenshot.

image

Probably because of `/r`.

You can check the API in https://api.jellyfin.org/openapi/jellyfin-openapi-stable.json.

liamnichols:
# Background

- Closes #44

It was reported that when a string that is used in comments uses `\r` for line breaks, the generated code will incorrectly fail to prefix the line with `///` resulting in non-compiling source files.

# Fix

From what I could tell, the easiest way to fix this is to separate into components based on anything within the predefined `.newlines` `CharacterSet`. 

It was a bit tricky to do this using `firstIndex(of:)` in the existing while loop, whereas it was much easier to use `components(separatedBy:)`. However, I am wary because of this comment:

```swift
// Unlike `components(separatedBy: "\n")`, it keeps empty lines.
```

However when I compared the implementation, I didn't see a difference in output:

Screenshot 2022-06-29 at 10 21 50

@kean, please could you confirm if I missed anything here? It looks like it works as expected using the updated approach but I might have misunderstood 🙏

### DIFF
--- a/Sources/CreateAPI/Generator/Templates.swift
+++ b/Sources/CreateAPI/Generator/Templates.swift
@@ -697,17 +697,8 @@ extension String {
 }
 
 extension String {
-    // Unlike `components(separatedBy: "\n")`, it keeps empty lines.
     var lines: [String] {
-        var lines: [String] = []
-        var index = startIndex
-        let input = self.trimmingCharacters(in: .whitespacesAndNewlines)
-        while let newLineIndex = input[index...].firstIndex(of: "\n") {
-            lines.append(String(input[index..<newLineIndex]))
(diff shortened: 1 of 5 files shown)


kean:
I missed the mention. Yes, I just tested it and it looks like it does keep the empty lines. I'm not sure I done it this way originally.

Issue #50, opened 2022-07-01
Update documentation


liamnichols:
I noticed that there are a few small inconsistencies with the documentation for the config in the README.md compared to the actual implantation.

Issue #48, opened 2022-06-29
Support Swift Package Plugins


liamnichols:
Swift 5.6 introduced the concept of Build Tool Plugins. Similar to Xcode build phases, they allow us to run tools as part of the build process.

Since CreateAPI is already nicely linked into the SPM ecosystem, we're in a place where its relatively easy for us to offer a plugin that can automatically generate Paths and Entities within a given target. 

We'd want to essentially enable people to create a **Package.swift** like this:

```swift
// swift-tools-version:5.6
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "MyPackage",
    platforms: [
        .macOS(.v10_15)
    ],
    products: [
        .library(name: "MyLib", targets: ["MyLib"]),
        .executable(name: "my-exec", targets: ["MyExec"])
    ],
    dependencies: [
        .package(url: "https://github.com/CreateAPI/CreateAPI.git", exact: "0.5.0"),
        .package(url: "https://github.com/kean/Get.git", exact: "0.8.0")
    ],
    targets: [
        .executableTarget(name: "MyExec", dependencies: ["MyLib"]),
        .target(
            name: "MyLib",
            dependencies: ["Get"],
            exclude: [
                "create-api.yaml",
                "schema.yaml"
            ],
            plugins: [
                .plugin(name: "CreateAPIPlugin", package: "CreateAPI")
            ]
        )
    ]
)
```

A manifest like the one above would then be capable of building CreateAPI (and Get) as part of the build process, reading from the **create-api.yaml** configuration and invoking the generator executable by providing the appropriate arguments. 

As a proof of concept, I created a simple initial implementation of `CreateAPIPlugin`:

```swift
@main
struct CreateAPIPlugin: BuildToolPlugin {
    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
        let config = target.directory.appending("create-api.yaml")
        let schema = target.directory.appending("schema.yaml")

        return [
            .buildCommand(
                displayName: "Generate with CreateAPI",
                executable: try context.tool(named: "create-api").path,
                arguments: [
                    "generate",
                    "--module", target.name,
                    "--config", config,
                    "--output", context.pluginWorkDirectory,
                    schema
                ],
                inputFiles: [
                    config,
                    schema
                ],
                outputFiles: [
                    context.pluginWorkDirectory.appending("Entities.swift"),
[...]

PR #52, opened 2022-07-02
Single source of truth for configuration options

liamnichols:
# Related Links

- #48 
- #50 

# Background

As it stands, the **GenerateOptions.swift** file contains a few things:

1. A series of types under the `GenerateOptionsSchema` class that are all optional allowing initialisation from an empty yaml/json file
2. A series of types under the `GenerateOptions` class and initialisers that map from the schema type with a default value (+ some additional customisations)

In addition to that, we have an example config file in the README document. 

I've identified this as a bit problematic and hard to maintain because we have two types to represent the same thing, the documentation is maintained manually (and a little out of sync) and the default values are also decoupled. 

I want to explore better ways to capture schema of the options with one single source of truth (that contains the properties, their defaults, and documentation). 

# Approach 

In this change, I'm reintroducing a new `ConfigOptions` type inside a new CreateOptions target (this is because it will also later be shared with the generator plugin target). This time, there are some differences: There is now just one single type where every property is defined, documented and has a default value. 

Now there are reasons why we didn't do this before:

1. `Decodable` ignores the default values defined for each property so an empty config would fail to deserialise
2. Some types needed to be mapped into something else that couldn't be represented in JSON/YAML before we passed it into the generator

But this is where I've chosen to use Sourcery to help out. 

Sourcery will generate two things:

1. **Docs/ConfigOptions.md** consisting of detailed documentation for every property
2. **Sources/CreateOptions/ConfigOptions+DecodableWithDefault.swift** consisting with a `Decodable.init(from:)` implementation that will fallback to the default value if it wasn't provided in the YAML/JSON

The caveat here is that whenever we modify **ConfigOptions.swift**, we have to re-run Sourcery (`make documentation`), but the upside there is that one type now acts as the single source of truth for everything. 

Usually I'd probably have given up and just dealt with the manual maintenance of options/docs/mapping, but since CreateAPI has a lot of flexibility and a rich set of options, I think it's important for this to be accurate. 

I then have a small wrapper around `ConfigOptions` that is called `GenerateOptions` (this will replace the old one). This wrapper currently just provides the small additional functionality of computing `allAcronyms` but could also be used in the future to abstract loading the config as well as tracking the file location in the event that we want to put file paths in the config files (so we can easily support relative paths).

In this PR, I've not started using the new `GenerateOptions` yet, that will come in a different PR. I'll also update the README in another PR too. 

While the PR is open, find the documentation here: https://github.com/CreateAPI/CreateAPI/tree/ln/options-and-documentation/Docs/ConfigOptions.md

I'd love any feedback on what can be improved with the format 🙏


Sources/CreateOptions/GenerateOptions+Acronyms.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        let allAcronyms = Self.defaultAcronyms
+            .union(addedAcronyms)
+            .subtracting(ignoredAcronyms)
+            .sorted { $0.count > $1.count }
+
+        Self.cache[hashValue] = allAcronyms
+        return allAcronyms
+    }
+}
```
liamnichols, line 28: I took this approach so that `GenerateOptions` could remain a struct, but now that I am thinking that it would be better to just keep `GenerateOptions` as a class instead. 

We have to be careful because we want to only compute this value once, but its accessed across multiple threads so this would have likely been a problem... I'll think of something better..

PR #61, opened 2022-07-27
Support automatically generating `Identifiable` conformance on entities

LePips:
As getting comfortable with the project, generate `Identifiable` if available on the entity. 

I initially had `Hashable` as well however that should be behind some config option because only structs have automatic conformance. I think  _all_ built in types can be statically listed and that hashable list is of course a subset of that list. Since `Hashable` would be desired in both `makeOneOf` and 'makeObject', for the future:
```swift
extension Array where Element == Property {
    var allHashable: Bool {
        let hashable = // list of all hashable types
        return entity.properties.allSatisfy { hashable.contains($0.type.builtinTypeName ?? "") }
    }
}
```

If this needs to be behind a config flag, I'm okay to do that.


Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -321,7 +321,14 @@ extension Generator {
         entity.properties = try makeInlineProperties(for: name, object: details, context: context)
             .filter { !$0.type.isVoid }
             .removingDuplicates(by: \.name) // Sometimes Swifty bool names create dups
-        entity.protocols = getProtocols(for: name, context: context)
+        
+        entity.protocols = {
+            var protocols = getProtocols(for: name, context: context)
+            let isIdentifiable = entity.properties.contains { $0.name.rawValue == "id" }
```
liamnichols, line 327: Are there likely to be common cases where the `type` of the `id` property isn't `Hashable`? 

I'm wondering if it would be preferable to check if the type is either `string` or `integer`? 

Maybe its better to start with a narrower scope that we are confident of, and if there are other types, we can widen it in the future.

liamnichols (commented):
Awesome! Thanks for the contribution @LePips! 

`Identifiable` is only available on iOS 13+ and while we don't officially state which OS version we try and maintain compatibility with, I know that some people use CreateAPI while still supporting iOS 12 (me 😬). 

I think putting this behind a flag is a good approach! Are you able to give that a go? 

Secondly, we'll want to make sure that this change is covered in some of our test fixtures so that nobody mistakenly breaks your change in the future 🙏 This process isn't very well documented at the moment, but either adapting an existing test or adding a new one to **GenerateOptionsTests.swift** should do the trick. 

You'll need to temporarily update the `projectPath` in **Snapshots.swift** and add a `GENERATE_SNAPSHOTS` environment variable that is set to `true` for it to update the expected fixtures. Let me know if you get stuck, I know its not the smoothest process at the moment 👍

LePips:
Oops, didn't think about the iOS versions. Thanks for the comment! I'll get on the tests.

Sources/CreateAPI/Generator/Generator+Schemas.swift
LePips: Good catch, didn't think about safety. Luckily, OpenAPIKit can tell us the subset of primitive types that it will decode to, mainly based upon OpenAPI 3: 

https://github.com/mattpolzin/OpenAPIKit/blob/main/Sources/OpenAPIKit/Schema%20Object/TypesAndFormats.swift

https://github.com/mattpolzin/OpenAPIKit/blob/main/Sources/OpenAPIKit/Schema%20Conformances/SwiftPrimitiveTypes+OpenAPI.swift

This is the small list I can create with all of the hashable types like I mentioned.

liamnichols:
Just one more thought, if somebody used the `rename` options in their config:

```yaml
rename:
  properties:
    User.user_id: id
    identifier: id
```

Would we be able to pick this up and annotate `User` as well as any other entity with the property `identifier` in their schema as `Identifiable`? 

I ask because I am not too familiar where the renaming takes place so we might need to adjust how we check?

LePips:
It will indeed attach `Identifiable` in that case with the replacement. The replacement of properties happen in L720 in `Generator+Schemas`, which happen when the entity properties are being set right before the protocols.

Sources/CreateAPI/Generator/Generator+Schemas.swift
liamnichols: I think we can probably save ourselves some trouble by only worrying about `string`, `integer` and maybe even `number` but I doubt anybody is seriously going to use `boolean`, `array` or `object` for an `id` property and expect it to also be `Identifiable`?

WDYT?

LePips: That seems like a safe assumption to make. We would then only have to worry about anything else if somebody requests it.

LePips: I attempted shortly to make a general `Hashable` conformance for all 4 `TypeIdentifier` types however it's not appropriate here. We can actually safely assume that `id` is a primitive type as all of the primitive types given to us conform to `Hashable`, either in initial declaration or an extension that is also `Encodable`, which all of the objects must be since they are parsed.

I only ran into an issue when it came to determining whether a user defined type was `Hashable` as that needs to recursively do other user defined types on properties based on the context. I'm sure there are other things to think about as well.

README.md
```diff
@@ -101,6 +101,8 @@ entities:
   baseClass: null
   # Protocols adopted by entities
   protocols: ["Codable"]
+  # Autogenerate Identifiable conformance if available
+  autogenerateIdentifiableConfomance: false
```
liamnichols, line 105: Please can we rename this to `isGeneratingIdentifiableConformance`. 

I'm not personally a huge fan of using Swifty naming conventions in yaml but it keeps it consistent with everything else 🙏

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -660,5 +660,45 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+                autogenerateIdentifiableConfomance: false
+            """, ext: "yaml")
+        ])
+        
+        // WHEN
+        try command.run()
+        
+        // THEN
+        try compare(package: "petstore-only-schemas")
+    }
```
liamnichols, line 683: No need for this one because the existing petstore schema tests would fail if the default value caused a regression 👍 

```suggestion
```

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -660,5 +660,45 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+            entities:
+                autogenerateIdentifiableConfomance: true
+            """, ext: "yaml")
+        ])
+        
+        // WHEN
+        try command.run()
+        
+        // THEN
+        try compare(package: "petstore-only-schemas")
```
liamnichols, line 702: Please can we generate this package as `petstore-identifiable` instead of `petstore-only-schemas` since it matches the test name. 

Also, just a note but with the current setup, we have to generate one package for each test we write here

liamnichols (commented):
Looking good!

I enabled the CI checks on this PR but they failed since we also need to commit the package that is generated by the tests. 

You can do that by setting the environment variable and running the tests locally:

Screenshot 2022-07-27 at 20 48 55

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -660,5 +660,25 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+    func testPestoreIdentifiableEnabled() throws {
+        // GIVEN
+        let command = try Generate.parse([
+            pathForSpec(named: "petstore", ext: "yaml"),
+            "--output", temp.url.path,
+            "--package", "petstore-only-schemas",
+            "--generate", "entities",
+            "--config", config("""
+            entities:
+                autogenerateIdentifiableConfomance: true
```
liamnichols, line 674: Could we also include the renaming in the example just to make sure that we have some coverage of that also

```suggestion
            entities:
                autogenerateIdentifiableConfomance: true
            rename:
                properties:
                    Error.code: id
```

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -321,7 +321,18 @@ extension Generator {
         entity.properties = try makeInlineProperties(for: name, object: details, context: context)
             .filter { !$0.type.isVoid }
             .removingDuplicates(by: \.name) // Sometimes Swifty bool names create dups
-        entity.protocols = getProtocols(for: name, context: context)
+        
+        entity.protocols = {
+            var protocols = getProtocols(for: name, context: context)
+            
+            if options.entities.isGeneratingIdentifiableConformance {
+                let isIdentifiable = entity.properties.contains { $0.name.rawValue == "id" && $0.type.isBuiltin}
```
liamnichols, line 329:
```suggestion
                let isIdentifiable = entity.properties.contains { $0.name.rawValue == "id" && $0.type.isBuiltin }
```

LePips:
Thanks for showing me the ropes!

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -321,7 +321,18 @@ extension Generator {
         entity.properties = try makeInlineProperties(for: name, object: details, context: context)
             .filter { !$0.type.isVoid }
             .removingDuplicates(by: \.name) // Sometimes Swifty bool names create dups
-        entity.protocols = getProtocols(for: name, context: context)
+        
+        entity.protocols = {
+            var protocols = getProtocols(for: name, context: context)
```
liamnichols, line 326: I had a quick scan, and it looks like `getProtocols(for:context:)` is used in four different locations since we create `EntityDeclaration`'s in a few different places. 

I tried to understand what impact that has, and it seems to be this.. Lets imagine the Petstore schema had this additional schema:

```diff
--- a/Tests/CreateAPITests/Specs/petstore.yaml
+++ b/Tests/CreateAPITests/Specs/petstore.yaml
@@ -97,6 +97,17 @@ components:
           type: string
         tag:
           type: string
+    Dog:
+      allOf:
+        - $ref: "#/components/schemas/Pet"
+        - type: object
+          properties:
+            breed:
+              type: string
+              enum:
+                - "Large"
+                - "Medium"
+                - "Small"
     Store:
       description:
       type: object
```

Because this is an `allOf` type, the entity is created differently, but when using `isInliningPropertiesFromReferencedSchemas`, it would still have an `id` and we'd therefore still want it to conform to `Identifiable` but it doesn't:

```diff
--- a/Tests/CreateAPITests/Expected/petstore-identifiable/Sources/Entities.swift
+++ b/Tests/CreateAPITests/Expected/petstore-identifiable/Sources/Entities.swift
@@ -21,6 +21,43 @@ public struct Pet: Codable, Identifiable {
     }
 }
 
+public struct Dog: Codable {
+    public var id: Int
+    /// Example: "Buddy"
+    public var name: String
+    public var tag: String?
+    public var breed: Breed?
+
+    public enum Breed: String, Codable, CaseIterable {
+        case large = "Large"
+        case medium = "Medium"
+        case small = "Small"
+    }
```

Looking at the way that the `Hashable` conformance is done too, it seems to suffer the same issue and needs to be fixed (we can do that separately 😄).

To fix it for `isGeneratingIdentifiableConformance` however, I think that we should move it inside of `getProtocols(for:context:)`. WDYT? 

I guess it means that we'll need to pass `[Property]` into the method as well, but that is fine. It will be useful for fixing the `Hashable` implementation later too (I think we can put that behind an option as well)

LePips: That makes a lot of sense, will do!

liamnichols: Thanks!

> Is there also a reason why `type` should be included as a parameter?

You're right, it looks unused so we can probably remove it 😄

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -321,16 +321,23 @@ extension Generator {
...
             .filter { !$0.type.isVoid }
             .removingDuplicates(by: \.name) // Sometimes Swifty bool names create dups
-        entity.protocols = getProtocols(for: name, context: context)
+        entity.protocols = getProtocols(for: entity, context: context)
+        
         return entity
     }
     
-    private func getProtocols(for type: TypeName, context: Context) -> Protocols {
+    private func getProtocols(for entity: EntityDeclaration, context: Context) -> Protocols {
```
LePips, line 329: I just went ahead and will pass in the entire entity, mainly because I _think_ we will have to check the nested types for `Hashable` declaration. I will take a look at that next and if not, I will change to just passing the properties there.

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -479,7 +486,7 @@ extension Generator {
         }.removingDuplicates { $0.type }
         
         entity.protocols = {
-            var protocols = getProtocols(for: name, context: context)
+            var protocols = getProtocols(for: entity, context: context)
             let hashable = Set(["String", "Bool", "URL", "Int", "Double"]) // TODO: Add support for more types
```
LePips, line 490: These will be removed when getting to `Hashable`

Tests/CreateAPITests/Specs/petstore.yaml
```diff
@@ -120,3 +120,14 @@ components:
...
+      allOf:
+        - $ref: "#/components/schemas/Pet"
+        - type: object
+          properties:
+            breed:
+              type: string
+              enum:
+              - "Large"
+              - "Medium"
+              - "Small"
```
liamnichols, line 133: I think that it's fine to remove this again now that we've verified that it works in that scenario that I described. Otherwise you'll have to regenerate all of the other fixtures that rely on **petstore.yaml** too.

liamnichols (commented):
Good job! Looking great! Just one more thing to get the tests passing again and then I think that its good to go 🚀

Tests/CreateAPITests/Specs/petstore.yaml
LePips: Sorry, I didn't see this before I committed, oops. If you'd like them removed I can revert.

LePips:
Sorry this PR has taken a bit of back and forth, I'm really appreciative of it and showing me how things should be done around here. I'll get the hang of it soon.

liamnichols (commented):
> Sorry this PR has taken a bit of back and forth, I'm really appreciative of it and showing me how things should be done around here. I'll get the hang of it soon. This project is great! We're going to be moving from OpenAPI after a few things are done.

No, thank you for taking the time to work on this! Your contribution is very valuable and it's good that we get to work though this 🙇 

I think we just need to revert that change still (reasoning inline) but then this is good to go 🎉 

Thanks again!

Tests/CreateAPITests/Expected/petstore-no-package/Entities.swift
```diff
@@ -39,6 +39,37 @@ public struct Error: Codable {
...
+        let values = try decoder.container(keyedBy: StringCodingKey.self)
+        self.pet = try Pet(from: decoder)
+        self.breed = try values.decodeIfPresent(Breed.self, forKey: "breed")
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var values = encoder.container(keyedBy: StringCodingKey.self)
+        try values.encode(pet, forKey: "pet")
+        try values.encodeIfPresent(breed, forKey: "breed")
+    }
```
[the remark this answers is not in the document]
kean: Great work on `Identifiable`. That's a nice addition for all SwiftUI users.

Yeah, `Dog`'s encoder is incorrect. I forgot to make it match the initializer.

```swift
         let values = try decoder.container(keyedBy: StringCodingKey.self)
         // This code reads properties from the existing decoder (not nested) 
         self.pet = try Pet(from: decoder)
         self.breed = try values.decodeIfPresent(Breed.self, forKey: "breed")
     }

         var values = encoder.container(keyedBy: StringCodingKey.self)
         // This should do the same: `pet.encode(to: encoder)`
         try values.encode(pet, forKey: "pet")
         try values.encodeIfPresent(breed, forKey: "breed")
     }
```

It would be nice to cover this piece with a bit uint test that checks that encoding/decoding pair actually works regardless of the implementation details.

liamnichols (approved):
Great work on this! Everything looks good to me 🚀 

I might work on the documentation improvements over the weekend and then get a release out after that. But if you want this in a tagged release before then, let me know!

PR #62, opened 2022-07-28
Tests - Automatic Path Finding and Remove Environment Variables

LePips:
`#filePath` is the closest we can get in regards to "automatic file path finding".

### DIFF
--- a/Tests/CreateAPITests/Snapshots.swift
+++ b/Tests/CreateAPITests/Snapshots.swift
@@ -26,7 +26,7 @@ func compare(expected: String, actual: String, file: StaticString = #file, line:
     }
 }
 
-fileprivate func diff(expectedURL: URL, actualURL: URL, file: StaticString = #file) throws {
+private func diff(expectedURL: URL, actualURL: URL, file: StaticString = #file) throws {
     func contents(at url: URL) throws -> [URL] {
         try FileManager.default.contentsOfDirectory(at: url, includingPropertiesForKeys: [.isDirectoryKey], options: [.skipsHiddenFiles])
     }
@@ -50,7 +50,7 @@ fileprivate func diff(expectedURL: URL, actualURL: URL, file: StaticString = #fi
 }
 
 @discardableResult
-fileprivate func shell(_ args: String...) -> Int32 {
+private func shell(_ args: String...) -> Int32 {
     let task = Process()
     task.launchPath = "/usr/bin/env"
     task.arguments = args
--- a/Tests/CreateAPITests/Snapshots.swift
+++ b/Tests/CreateAPITests/Snapshots.swift
@@ -5,8 +5,8 @@
 import Foundation
 import XCTest
 
-fileprivate let generateSnapshots = false
-fileprivate let openDiff = false
+private let generateSnapshots = false
+private let openDiff = false
 
 func compare(expected: String, actual: String, file: StaticString = #file, line: UInt = #line) throws {
     let expectedURL = Bundle.module.resourceURL!
--- a/Tests/CreateAPITests/Snapshots.swift
+++ b/Tests/CreateAPITests/Snapshots.swift
@@ -5,27 +5,28 @@
 import Foundation
 import XCTest
 
+fileprivate let generateSnapshots = false
+fileprivate let openDiff = false
+
 func compare(expected: String, actual: String, file: StaticString = #file, line: UInt = #line) throws {
-    let env = ProcessInfo.processInfo.environment
     let expectedURL = Bundle.module.resourceURL!
         .appendingPathComponent("Expected")
         .appendingPathComponent(expected)
     let actualURL = URL(fileURLWithPath: actual)
     
-    if env["GENERATE_SNAPSHOTS"] == "true" || !FileManager.default.fileExists(atPath: expectedURL.path) {
-        let testsFolderPath = #filePath
-            .split(separator: "/")
-            .dropLast()
-            .joined(separator: "/")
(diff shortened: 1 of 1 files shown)


Tests/CreateAPITests/Snapshots.swift
```diff
@@ -5,26 +5,27 @@
...
         .appendingPathComponent(expected)
     let actualURL = URL(fileURLWithPath: actual)
+    
     if env["GENERATE_SNAPSHOTS"] == "true" || !FileManager.default.fileExists(atPath: expectedURL.path) {
-        let projectPath = (projectPath as NSString).expandingTildeInPath
-        let destinationURL = URL(fileURLWithPath: projectPath + "/Tests/CreateAPITests/Expected/\(actualURL.lastPathComponent)")
+        let testsFolderPath = #filePath
+            .split(separator: "/")
+            .dropLast()
+            .joined(separator: "/")
```
LePips, line 19: If this functional indentation style is not preferred, happy to inline.

LePips:
Is there a specific reason why an environment variable is set to generate versus a static bool? I think it would be more convenient if it was just a static bool for general development and then a `FORCE_GENERATE` environment would override that to force generation.

Tests/CreateAPITests/Snapshots.swift
liamnichols: > If this functional indentation style is not preferred, happy to inline.

I like it 🙂 

> Speaking of style, should we make a lint process for us?

Great idea 

- #64

Tests/CreateAPITests/Snapshots.swift
```diff
@@ -5,26 +5,27 @@
...
+        let destinationURL = URL(fileURLWithPath: "/" + testsFolderPath + "/Expected/\(actualURL.lastPathComponent)")
```
liamnichols, line 20: I'm just wondering if we can go one step further and create `URL(fileURLWithPath: #filePath)` and then use `appendingPathComponent(_:)`/`deletingLastPathComponent()` to construct the final path instead of relying on string manipulation directly? 

```swift
let destinationURL = URL(fileURLWithPath: #filePath)
    .deletingLastPathComponent()
    .appendingPathComponent("Expected")
    .appendingPathComponent(actualURL.lastPathComponent)
```

liamnichols (commented):
Oh awesome! Thanks so much for picking this up 🙌 

> Is there a specific reason why an environment variable is set to generate versus a static bool? I think it would be more convenient if it was just a static bool for general development and then a `FORCE_GENERATE_SNAPSHOT` environment variable would override that to force generation elsewhere.

Not that I'm aware of. I agree though, a static bool is easier to conditionally enable I think so happy to change it (same with the `OPEN_DIFF` var)

LePips:
I'll be able to provide steps for testing after, or during, your README update PR as mentioned in your current PR https://github.com/CreateAPI/CreateAPI/pull/52

Tests/CreateAPITests/Snapshots.swift
```diff
@@ -5,26 +5,28 @@
 import Foundation
 import XCTest
 
-// TODO: Find it automatically
-var projectPath = ("~/Developer/CreateAPI/" as NSString).expandingTildeInPath
+fileprivate let generateSnapshots = false
+fileprivate let openDiff = false
```
LePips, line 9: This or UPPER_CASE?

Tests/CreateAPITests/Snapshots.swift
```diff
@@ -5,26 +5,28 @@
...
         try? FileManager.default.removeItem(at: destinationURL)
         try FileManager.default.copyItem(at: actualURL, to: destinationURL)
     } else {
-        try diff(expectedURL: expectedURL, actualURL: actualURL, file: file, line: line)
+        try diff(expectedURL: expectedURL, actualURL: actualURL, file: file)
     }
 }
 
-private func diff(expectedURL: URL, actualURL: URL, file: StaticString = #file, line: UInt = #line) throws {
+fileprivate func diff(expectedURL: URL, actualURL: URL, file: StaticString = #file) throws {
```
liamnichols, line 29: What's the need to change these to `fileprivate` when `private` does the trick? 

https://realm.github.io/SwiftLint/private_over_fileprivate.html

Tests/CreateAPITests/Snapshots.swift
```diff
@@ -48,7 +50,7 @@ private func diff(expectedURL: URL, actualURL: URL, file: StaticString = #file,
 }
 
 @discardableResult
-func shell(_ args: String...) -> Int32 {
+fileprivate func shell(_ args: String...) -> Int32 {
```
liamnichols, line 53: Lets make this `private` too

Tests/CreateAPITests/Snapshots.swift
liamnichols: Looks good how it is! 👍

liamnichols (commented):
Looking great! Thanks for sorting the environment variables too 🙏 

I think we should prefer `private` over `fileprivate` when at the top-level but other than that, everything seems great 👍

Tests/CreateAPITests/Snapshots.swift
LePips: I guess just a personal preference, where `fileprivate` is more declarative that it's, well, private to the file. Doesn't really matter to me though.

liamnichols (approved):
Awesome stuff! Thanks for doing this, it makes it so much easier to test locally 🙇

Issue #69, opened 2022-08-01
Homebrew Installation


LePips:
I rather prefer Homebrew to install all my packages instead over Mint and of course over manual installation for every update. Could there be a Homebrew option for installation?

liamnichols:
Generally speaking, this would be a good idea especially because it would help improve adoption of the tool. 

--- 

Out of curiosity for your specific use case though, is a Swift Package Plugin something that might work out well for you? You can check out one implementation here: AvdLee/appstoreconnect-swift-sdk#191 

I'm also actively looking at other plugin-based workflows that might be useful for CreateAPI users over in #48.

LePips:
I personally haven't used a Swift Package Plugin yet but definitely take a look at it!

LePips:
Additionally, I think a `--version` flag needs to be added.

liamnichols:
> Additionally, I think a `--version` flag needs to be added.

👍 Yep that would be useful! I think swift-argument-parser has support built in, we just need to set it on the `CommandConfiguration` iirc

liamnichols:
0.0.5 is now available via Homebrew! Thanks for the suggestion 

https://formulae.brew.sh/formula/create-api

```
➜ brew install create-api
==> Downloading https://ghcr.io/v2/homebrew/core/create-api/manifests/0.0.5
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/create-api/blobs/sha256:99d2635
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sh
######################################################################## 100.0%
==> Pouring create-api--0.0.5.arm64_monterey.bottle.tar.gz
🍺  /opt/homebrew/Cellar/create-api/0.0.5: 6 files, 9.1MB
==> Running `brew cleanup create-api`...
Disable this behaviour by setting HOMEBREW_NO_INSTALL_CLEANUP.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
➜ create-api generate --help
USAGE: create-api generate [<options>] <input>
```

PR #73, opened 2022-08-01
Add LICENSE

liamnichols:
- #69 

To publish to Homebrew, all projects must have a license and I noticed that most repos in the CreateAPI org were missing one. I'm following what HTTPHeaders uses and going with MIT. 

If you are ok with this @kean, I'll go ahead and just push the file directly to the other repos rather than creating Pull Requests for each one.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Alexander Grebenyuk
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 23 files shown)


liamnichols:
I missed the fact that most of the source files already referenced the MIT license in the header. I've opted to remove the headers in favour of just the single **LICENSE** file instead.

LePips:
Does the org have a copyright?

kean:
Yes, everything goes under MIT. Please go ahead, @liamnichols.

kean (approved):

liamnichols:
> Does the org have a copyright? Or, should copyright be included in a header?

The CreateAPI org currently is just a GitHub org for grouping the CreateAPI related repositories in one place so has no kind of copyright afaik

Issue #64, opened 2022-07-28
Setup a code linter


liamnichols:
@LePips raises a good point that we should probably simplify some things around code review by setting up some basic rules with a linting tool. 

I'm suggesting SwiftLint since I'm familiar with it, but open to other options too. 

We should be able to run it locally and also it should run as a check on Pull Requests.

LePips:
I guess to get started, what rules do we want? A first `autocorrect` run doesn't produce much, only `Array<...>` to `[...]` and some whitespace removed.

liamnichols:
That sounds like a good start. If we can start with the default rules that are enabled and correct everything then that is probably already good enough.

Maybe we can take some examples from other public style guides beyond that? Some that I know of:

- https://github.com/airbnb/swift/blob/master/Sources/AirbnbSwiftFormatTool/swiftlint.yml
- https://github.com/raywenderlich/swift-style-guide/blob/main/com.raywenderlich.swiftlint.yml

LePips:
I started with the raywenderlich config and just removed some options to get to this config: https://github.com/LePips/CreateAPI/blob/swiftlint/.swiftlint.yaml

You can see what changes it made with a first autocorrect: https://github.com/LePips/CreateAPI/commit/80551e91b57d8c2ac1dbe984827615ea38be236a

Overall, mainly fixes spacing and changes `Array<type>` declarations to `[type]`.

liamnichols:
Looks like a good start! 

We’ll want to ignore the redundant default value in ConfigOptions.swift because that is used to generate documentation, and we’ll also want to ignore the DecodableWithDefault extension since that’s generated code

LePips:
Disabled the redundant rule for ConfigOptions.swift and all rules for the generated files: https://github.com/CreateAPI/CreateAPI/commit/91af76e716342b5af746048029eac46a6f0bbf87

PR #79, opened 2022-08-02
Lint project using Swfitlint

LePips:
- https://github.com/CreateAPI/CreateAPI/issues/64

Initial `swiftlint autocorrect` but with no CI. Config rules are up for discussion.


LePips:
The last file change seems odd, unless the test needed to be updated elsewhere and wasn't caught then.

Sources/CreateAPI/Generator/Templates.swift
```diff
@@ -316,17 +317,17 @@ final class Templates {
...
         """
-        
+
         return """
         \(access)init(from decoder: Decoder) throws {
-            
+
             struct Discriminator: Decodable {
                 let \(discriminator.propertyName): String
             }
```
liamnichols, line 333: This is what caused the change to generated code @LePips. 

I don't think it's an issue since its just whitespace (better to strip it though) but its worth being mindful that the linter might mistakenly mess some other stuff up.

liamnichols (approved):
Looks good to me! Thanks for picking this up

liamnichols:
I'll merge this now since there are minimal actual changes, we can discuss adding/adjusting rules later on 👍

Issue #58, opened 2022-07-25
Any docs on how to actually use the generated code?


soemarko:
I am just lost on how to actually consume my generated API.

What I've done is generate the the `Entities.swift` and `Paths.swift` drag an drop it into the project. Add Get and URLQueryEncoder via SPM. But now I'm faced with a bunch of `Cannot find type 'api' in scope`

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/381585/180772309-22a77cfb-69ee-4cb3-8346-baf84c993c3e.png">

Any info on what to do next would be nice. Thank you!

soemarko:
Figured it out. Should've generated the code with package instead of module. And to help others that are facing the same wall here's what I did;

`$ create-api generate --output api --package api swagger.json`

Then from SPM add local package. It'll also pull Get and URLQueryEncoder as well.

Next to quickly test consuming it:
```
		let client = APIClient(baseURL: URL(string: "https://api.server.com"))
		
		Task.init(operation: {
			do {
				let response = try await client.send(api.Paths.banners.get)
				let banners = response.value.result ?? []
				
				for banner in banners {
					print(banner.imageURL)
				}
			}
			catch let error {
				print(error)
			}
		})
```

liamnichols:
Thanks for the feedback @soemarko! I'll actually reopen this because a Getting Started guide would absolutely be useful to help in this kind of scenario 👍

soemarko:
> Thanks for the feedback

No problem. Looking further into this, I believe most people come here from Get, so it would be easier for them. But I came here from [SwagGen](https://github.com/yonaskolb/SwagGen). So yeah, it'll be great to have some kind of docs.

Rusik:
I wanted to play with it and faced exactly the same issue, couldn't figured out or found any docs how to actually use generated code

mgrider:
I'll chime in and say a list of projects that use this somewhere (maybe even in the README) would also be quite useful. (I hadn't seen that `appstoreconnect-swift-sdk` project before!)

Issue #80, opened 2022-08-02
`--package` should use Get starting from version 1.0.0


liamnichols:
Get v1 was released recently, our generated Package.swift files should start from this version instead because they won't automatically bump to the next major version.

Issue #42, opened 2022-06-15
requestOperationIdExtension adds private extension making it inaccessible if in a separate file


andrea-gradecak:
In [Templates.swift](https://github.com/CreateAPI/CreateAPI/blob/5225618ac7fd542fcb89d27413d568c505fa912f/Sources/CreateAPI/Generator/Templates.swift#L622), `requestOperationIdExtension` adds a private extension.

If generation creates separate files, the private extension is also in its own file `Paths+Extensions` which makes it inaccessible to use in other files.

LePips:
Just taking a look at this, should this extension by added by default and we always generated a functional helper function:
```swift
public extension Request {
    public func settingID(_ id: String) -> Request {
        var copy = self
        copy.id = id
        return copy
    }
}
```

liamnichols:
Could it even just be a case of removing the `private` from the existing template?

LePips:
Or, this could be a change in Get itself as it seems more appropriate there

liamnichols:
Using `isAddingOperationIds` will fail if the schema doesn't specify an `operationId` for one or more of its operations, which is allowed as per the spec. 

So at the moment, the flag is necessary. But that being said, I've seen other tools automatically infer the operationId from the path + method, which we could also do. And if we did, then we'd be able to just generate it automatically (or maybe just leave it as `nil`?) 

> Or, this could be a change in Get itself as it seems more appropriate there

Yeah I did think about that and it is also a possibility. It would simplify things

LePips:
I think it would be safe to add the `operationId` as the `Request.id` by default, `nil` if none specified. Get does not use the `id` property so it's mainly for developer usage (unless I'm missing something).

liamnichols:
Before I noticed the flag, I asked about the purpose in kean/Get#33 and you are right, it is designed for this. 

So in that case, maybe we do just pass the `id` through the convenience methods.. something like this:

```swift
public static func get(_ url: String, query: [(String, String?)]? = nil, headers: [String: String]? = nil, id: String? = nil) -> Request {
    Request(method: "GET", url: url, query: query, headers: headers, id: id)
}
```

What do you think about this @kean?

kean:
[...]

PR #82, opened 2022-08-03
Update README and add Getting Started Tutorial, Swift Package Plugin docs, CONTRIBUTING guidance + more

liamnichols:
# Background 

- closes https://github.com/CreateAPI/CreateAPI/issues/58
- closes https://github.com/CreateAPI/CreateAPI/issues/48

In the past, we never invested much time documenting the tool and best practices around using it. To help with adoption of all the awesome features in CreateAPI, we want to make sure that people now how to use them and the aim here is to do that via plenty of documentation.

# Changes

To get an overall feel for the changes, check out the repo here: https://github.com/CreateAPI/CreateAPI/tree/ln/documentation

In particular, I've done the following:

- Rewritten [README.md](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/README.md)
- Added [CONTRIBUTING.md](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/CONTRIBUTING.md)
- Written a tutorial: [Generating an API with CreateAPI](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/Docs/Tutorial.md)
- Documented [Swift Package Plugins](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/Docs/SwiftPackagePlugins.md)
- Added some notes about [Dependencies](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/Docs/Dependencies.md)
- Kick started an [Advanced Setup](https://github.com/CreateAPI/CreateAPI/blob/ln/documentation/Docs/AdvancedSetup.md) doc with a guide to "Using a different API Client"

Some of the contents in this PR rely on tweaks that are still upcoming for the 0.1 release, so it might not quite match to how CreateAPI is used today.. I'll add notes inline.


CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+If you are making a change, ether to fix/reproduce a bug, or to add a new feature, you can enable some helpers in **Snapshots.swift** to help you:
+
+1. Set `generateSnapshots` to `true` in order to export your changes to the Expected fixtures.
+2. Set `openDiff` to `true` to view a rich diff of the generated output vs the expected output. This is useful whe debugging.
+
+Please be sure not to commit changes to these flags when submitting a Pull Request.
+
+If you want to add a new spec or configuration to the tests, you can add the schema to **Tests/CreateAPITests/Specs/** and then write a new test, with the desired configuration in **GenerateTests.swift**. Be sure to enable `generateSnapshots` on the first run.
+
+In addition if you've generated a package, you should include it in **Tests/CreateAPITests/AllPackages/Package.swift** so that CI checks can confirm that it compiles ok on both macOS and Linux.
```
liamnichols, line 43: Only accurate after #81

Docs/AdvancedSetup.md
```diff
@@ -0,0 +1,45 @@
...
+```swift
+struct Request<Response> {
+    var method: String = "GET"
+    var url: String
+    var query: [(String, String?)]?
+    var body: Encodable?
+    var headers: [String: String]?
+    var id: String?
+}
+```
```
liamnichols, line 43: This would only work when we tweak how the generated Paths use `Request` (see discussion in #42)

Docs/Dependencies.md
```diff
@@ -0,0 +1,15 @@
+# Dependencies
+
+By default, CreateAPI will use [Get](https://github.com/kean/Get) for the API client. If you are generating a package, this dependency is setup for you automatically but if you are generating for an existing module, you will need to manage dependencies yourself.
+
+Below is a table that describes dependencies that CreateAPI generated code might need to use:
+
+Dependency|Minimum Version|Required When?
+---|---|---
+[Get](https://github.com/kean/Get)|1.0.0|Generating Paths*
```
liamnichols, line 9: Currently its 0.3.1, but I'll bump it to 1.0.0 as part of #80

Docs/OpenAPISupport.md
```diff
@@ -0,0 +1,12 @@
...
+The goal is to completely cover OpenAPI 3.x spec.
+
+Currently, the following features are **not** supported:
+
+- External References
+
+Some discrepancies with the OpenAPI spec are by design:
+
+- `allowReserved` keyword in parameters is ignored and all parameter values are percent-encoded
+- `allowEmptyValue` keyword in parameters is ignored as it's not recommended to be used
```
liamnichols, line 12: This was in the README.md, I moved it here but didn't change it's contents (other than removing "upcoming")

Docs/Tutorial.md
```diff
@@ -0,0 +1,146 @@
...
+
+And you will find a new folder called **PetstoreKit** located within the output directory. Double click on the **Package.swift** file within the package to open it in Xcode and you will see something like the following:
+
+<img
+  src="./Images/Tutorial_01_NewPackage.png#gh-light-mode-only"
+  alt="A screenshot of Xcode focusing on the Paths.swift source file within the newly generated PetstoreKit package. To the left, you can see the project navigator which displays the source files (Entites.swift, Paths.swift) and the resolved dependencies (Get, HTTPHeaders, URLQueryEncoder)."
+/><img
+  src="./Images/Tutorial_01_NewPackage_Dark.png#gh-dark-mode-only"
+  alt="A screenshot of Xcode focusing on the Paths.swift source file within the newly generated PetstoreKit package. To the left, you can see the project navigator which displays the source files (Entites.swift, Paths.swift) and the resolved dependencies (Get, HTTPHeaders, URLQueryEncoder)."
+/>
```
liamnichols, line 48: I went all out on this 😅 

https://user-images.githubusercontent.com/482871/182634333-0d1d78f2-5b50-40a2-b197-b1c4570fc55c.mp4

https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

README.md
```diff
@@ -75,25 +87,46 @@ OPTIONS:
...
+    > Checksum: `89c75ec3b2938d08b961b94e70e6dd6fa0ff52a90037304d41718cd5fb58bd24`
+    >
+    > ```swift
+    > .binaryTarget(
+    >     name: "create-api",
+    >     url: "https://github.com/CreateAPI/CreateAPI/releases/download/0.0.5/create-api.artifactbundle.zip",
+    >     checksum: "89c75ec3b2938d08b961b94e70e6dd6fa0ff52a90037304d41718cd5fb58bd24"
+    > )
+    > ```
+- [ ] Push the release to Homebrew with `brew bump-formula-pr create-api`.
```
liamnichols, line 132: I had been meaning to note this down somewhere, maybe README.md isn't the best place for it though?

liamnichols, line 132: Depends on Homebrew/homebrew-core#107160

liamnichols:
cc @soemarko, @Rusik, @mgrider, @LePips, @kean... Any feedback on this will be much appreciated, thanks!

Docs/Tutorial.md
```diff
@@ -0,0 +1,146 @@
...
+
+Once the dependencies have resolved, build the project and everything will succeed. Congratulations, you've generated your first project using CreateAPI!
+
+## Using the Code
+
+Now that you have the generated Paths and Entities, you probably want to actually put them to good use!
+
+In the **Sources** directory, add a new file called **PetstoreClient.swift**. Within that file, add the following code:
```
mgrider, line 56: I kind of think you should add a note here about how you don't have to modify the generated package, you can instead include it in your own project. For APIs that are in development simultaneously to the Swift project, you'll probably be generating the package over and over again during development, so modifying it directly is probably not a great idea.

This leads me to my next thought, which is maybe there could be a "best practices" document with some tips and tricks in it? A couple of ideas for that:

1. Put your CreateAPI generation command in a bash script that's checked into git somewhere, (along side the generated package), so it's easy for anyone on the project to update the API by running the script.
2. A note about writing tests that check for all the properties in the generated package that you are using specifically, so when the API is updated you can run your tests and make sure the API still gives you back the objects and properties you are using in your application.

mgrider: I see that you do address overwriting changes to the generated package when talking about modules. I still think a note about using the package as opposed to modifying it here would be nice.

README.md
mgrider: Agree that maybe the README isn't the best fit for this. I thought about suggesting you move it to `CONTRIBUTING.md`, but maybe a better option is a new file entirely. You could still link to it from the readme, or from contributing.

mgrider (approved):
I had some suggestions, but overall I really love the improvements here. A+ work!

Docs/AdvancedSetup.md
kean: I added `imports: ["Get"]` parameter in the initial version as a basic way to exclude `Get` dependency. But the goal was to eventually provide a higher-level API for that, but I never got to actually adding it. If you are using a custom client, the `Request` type should probably be auto-generated by CreateAPI. The user could than provide extensions to their networking code to handle `Request`.

Docs/README.md
```diff
@@ -0,0 +1,10 @@
+# Documentation
+
+Here you can find documentation specific for using CreateAPI.
+
+- [Tutorial: Generating an API with CreateAPI](./Tutorial.md)
```
kean, line 5: Have you considered generating the documentation using DocC?

Docs/Tutorial.md
kean: Love this.

Docs/Tutorial.md
kean: Since package is the default/recommended option (AFIAR), ideally you should never modify its contents. One of the main initial goals was to provide enough tools to generate code that compiles and you are happy with, making this workflow (always autogenerate) feasible.

kean: There is also a `--watch` option (for autogeneration) which I'm not sure is documented.

kean (commented):
Impressive work on the docs!

README.md
LePips: I think linked from contributing would be the way, especially since releases won't be made unless you have access to.

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+Submitting Pull Requests is always welcome, this could be to fix bugs, refactor code, implement new features or even to add or update documentation.
+
+If you are changing code, please be sure to make sure that the tests pass locally before submitting your Pull Request. You can do this by running `swift test` in terminal (or <kbd>⌘</kbd> + <kbd>U</kbd> in Xcode).
+
+### Test Fixtures
+
+The CreateAPITests target has a series of expected fixtures committed to the repository and when running tests, to verify that the generator works as expected, the output generated by the test must match the contents committed to the repo.
+
+If you are making a change, ether to fix/reproduce a bug, or to add a new feature, you can enable some helpers in **Snapshots.swift** to help you:
```
LePips, line 34: Pedantic change, but I think files would be in `code` because **bold** can be hard to see.

LePips:
Looks great!

CONTRIBUTING.md
liamnichols: heh yeah, I guess it depends a bit on the importance but I follow the Ray Wenderlich formatting guidelines (they're not public I don't think, but I occasionally tech edit there) when I use bolding for filenames. 

I'll stick with using bold because I don't think there is too much importance on the filenames in here 🙇

Docs/README.md
liamnichols: It crossed my mind, but I haven't looked into it too much. Something totally worth doing in the future though once we've build up and stabilised the docs I think 👍

Docs/AdvancedSetup.md
liamnichols: > I added `imports: ["Get"]` parameter in the initial version as a basic way to exclude `Get` dependency. But the goal was to eventually provide a higher-level API for that, but I never got to actually adding it. If you are using a custom client, the `Request` type should probably be auto-generated by CreateAPI. The user could than provide extensions to their networking code to handle `Request`.

Good to know your thinking. I only document this as it's how we're approaching it in our project but I do agree, the current approach isn't ideal. 

I imagine that a `isUsingGet` (or `isUsingCustomAPIClient`) flag similar to `isGeneratingResponseHeaders` and `isNaiveDateEnabled` might be a good idea in the short term. Including `Request` in the generation is also a good idea, I didn't think about that (also I didn't realise that the struct was so simple until I added it here 😄)


> I can see people not desiring a `Request` object at all, using their CreateAPI generation for only the objects and string paths. _Not that I would do it myself_ but this seems safer to provide so that we don't "force" a dev to have a `Request` shim at all. However, this would be a big change for more discussion.

Interesting.. Although how about all of the other information (`method`, `query`, `body` etc?) I think `Request` is pretty lightweight on its own but open to other options. One of the less obvious benefits of `Request` is that it uses generics to associate the appropriate response type and I think people find that valuable?

README.md
liamnichols: Great idea, I'll move it to **CONTRIBUTING.md**, thanks

Docs/AdvancedSetup.md
LePips: I just thought of that use case so it might not even be valid or something completely out of scope. I really like kean's original idea of CreateAPI defining `Request` itself and then devs extend off of that.

Docs/Tutorial.md
liamnichols: So the aim of from the tutorial was to demonstrate both approaches to help you understand what is possible. It wasn't necessarily to say "you must do it like this" but I guess that I can make that a bit clearer 👍

>  I still think a note about using the package as opposed to modifying it here would be nice.

I'll do this then 👍 

> Since package is the default/recommended option (AFIAR), ideally you should never modify its contents. One of the main initial goals was to provide enough tools to generate code that compiles and you are happy with, making this workflow (always autogenerate) feasible.

Yup that's cool.

From trying things out, I guess the current issue I found with how CreateAPI works today is that you can't add extra targets or tests. The idea of having CreateAPI generate the whole package is great, but it means that the rest of your project has to go in a separate project/package where traditionally you might have just had them as two targets in the same package.  

For example, I personally use CreateAPI today in `--module` mode to integrate it into my existing networking client which has to do a lot of other stuff (mostly related to auth). [appstoreconnect-swift-sdk](https://github.com/avdLee/appstoreconnect-swift-sdk) also does something very similar. 

I'll add a note, but I agree that it would be great to revisit this in the future once we've built on the current capabilities somewhat. 

For example, I had an shower thought that maybe we could benefit from a `create-api init` command that offers a guided process to setup a package (repo), pull the schema into the right location, set a default config and even integrate a Swift Package Plugin to enable easy regeneration each time its required 😬

kean: > in a separate project/package where traditionally you might have just had them as two targets in the same package

That just goes to show again that as someone who is not actively using the tool, it was a great idea to ask for help to maintain it. I was thinking about a scenario where you have an Xcode Project, you generate the package in one of the subfolders, and then integrate it as a local dependency.

liamnichols: > I was thinking about a scenario where you have an Xcode Project, you generate the package in one of the subfolders, and then integrate it as a local dependency.

To be honest, it does sound like a perfect use case too so I think it's a great idea supporting both. The majority of my SPM experience comes from outside of macOS/iOS Xcode Projects so I haven't thought much about this before. 

---

I plan to add the above disclaimer to the document:

```suggestion
> **Note**: In some cases, you might prefer to use the generated Package as-is by referencing it as a dependency in another project or package (this allows you to regenerate it whenever the spec changes), but for the educational purposes in this tutorial you will work directly inside the package instead.

In the **Sources** directory, add a new file called **PetstoreClient.swift**. Within that file, add the following code:
```

mgrider: > I was thinking about a scenario where you have an Xcode Project, you generate the package in one of the subfolders, and then integrate it as a local dependency.

This is definitely how I'm using it currently. Personally, I love that it's a separate package and import.

liamnichols: Good to know thank you. I think there certainly is still room for another document that lists out various different possibilities explaining how you can go about using CreateAPI. It'll even be useful for me/other maintainers 😄 

I'll probably leave it out of this PR for now, but might come back to it after if somebody else doesn't beat me to it.

README.md
```diff
@@ -16,30 +16,42 @@ Delightful code generation for OpenAPI specs for Swift written in Swift.
...
 
 ```bash
-mint run CreateAPI create-api generate -h
+$ brew install create-api
 ```
 
+### Swift Package Plugins
+
+- [Creating a Swift Package Plugin](./Docs/SwiftPackagePlugins.md)
+
```
LePips, line 31: Now that we're available through Homebrew, should have installation instructions.

LePips (changes requested):

README.md
LePips: Clicked the wrong line, woops.

liamnichols: It's just above this line already 😄

LePips: Oh, apologies I missed it.

LePips (approved):

Docs/Tutorial.md
liamnichols: So I've actually revisited this when updating everything to reflect the changes in 0.1.0. I decided to focus the tutorial entirely on generating packages and make no mention of the alternative approach. 

Later on I will add to the **AdvancedSetup.md** with instructions on how to integrate into an existing project 👍

mgrider: I went and took a look at your new Tutorial, and it's _fantastic_! Really excellent work! 🤩

PR #83, opened 2022-08-03
[BREAKING] Require Get 1.0.0 or greater

### LINKED ISSUE #42: requestOperationIdExtension adds private extension making it inaccessible if in a separate file

### LINKED ISSUE #80: `--package` should use Get starting from version 1.0.0

liamnichols:
# Background 

- Closes #42 
- Closes #80

Get recently released a new major version 1.0.0 (and 1.0.1), meaning that current CreateAPI generated packages will not use anything beyond 0.8.0 as per the semantic versioning requirements defined in the generated **Package.swift** files. 

To coincide with CreateAPI's 0.1.0 release, there are also some semi-breaking enhancements in this area that we can make to help fix some other related issues and minimise the dependence on implementation details of Get. 

# Changes 

In this pull request, I not only bump the version defined in the generated **Package.swift** to 1.0.0, but I also update the Paths generator to stop using the telescoping methods in favour of using the `Request` initialiser directly. 

By doing this, it greatly simplifies the requirements of the `Request` type meaning that Get is free to change other aspects of its interface and CreateAPI is free to offer a simple `Request` type when the user does not want to depend on Get directly. 

This results in diffs to the generated code like the following:

```diff
 public func patch(_ body: edgecases_tabs.Client) -> Request<edgecases_tabs.Client> {
-    .patch(path, body: body)
+.   Request(method: "PATCH", url: path, body: body)
 }
```

I was also able to cleanup the (sometimes faulty) `requestOperationIdExtension` since the `id` parameter is available on the initialiser. 

# Implications

This change comes with two _potentially_ breaking implications to be aware of:

1. If you use Get, you'll need to migrate from 0.x.x to 1.x.x
2. If you didn't use Get and had been providing your own `Request` type, you might need to make sure that you expose the appropriate initialiser* 

As far as I am aware, there are no major breaking changes in Get. The release notes can be found here: https://github.com/kean/Get/releases/tag/1.0.0

### DIFF
--- a/Sources/CreateAPI/Generator/Generator+Package.swift
+++ b/Sources/CreateAPI/Generator/Generator+Package.swift
@@ -3,7 +3,7 @@ import Foundation
 extension Generator {
     func makePackageFile(name: String) -> String {
         let packages: String = [
-            #".package(url: "https://github.com/kean/Get", from: "0.3.1")"#,
+            #".package(url: "https://github.com/kean/Get", from: "1.0.2")"#,
             isHTTPHeadersDependencyNeeded ? #".package(url: "https://github.com/CreateAPI/HTTPHeaders", from: "0.1.0")"# : nil,
             isNaiveDateNeeded ? #".package(url: "https://github.com/CreateAPI/NaiveDate", from: "1.0.0")"# : nil,
             isQueryEncoderNeeded ? #".package(url: "https://github.com/CreateAPI/URLQueryEncoder", from: "0.2.0")"# : nil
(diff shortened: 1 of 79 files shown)


kean:
I'm working on addressing the design issues with the `Request` that I raised earlier. Here's a [draft MR](https://github.com/kean/Get/pull/51) that I would love to get your and @LePips feedback on.

I believe it won't affect the existing CreateAPI implementation too much.

```swift
// Before
Request(method: "PATCH", url: path, body: body)

// After
Request(method: .patch, path: path, body: body) 
```

PR #86, opened 2022-08-04
Replace `entities` option `isGeneratingCustomCodingKeys` with `isUsingStringsForCodingKeys`

### LINKED ISSUE #75: Enable `isGeneratingCustomCodingKeys` by default
# Background

- #57 

# Changes 

Generating a `CodingKey` conforming enum for every entity has a significant impact on the output binary size, even after optimisation and especially in larger projects. 

It's much more efficient to leverage a single `CodingKey` type when working with generated code, which is what happens when `isGeneratingCustomCodingKeys` is set to `false`. 

Essentially, this should be the default behaviour and generating custom keys should be an opt in feature (or maybe even just removed).

To ease a transition, we might well want to rename the option too:

> The name `isGeneratingCustomCodingKeys` is also a bit poor. It should probably just spell out the behavior: `isUsingRawStringsAsCodingKeys` or something like that.

liamnichols:
# Background

- closes #75 
- #57 

Using individual enums for `CodingKey` types can unexpectedly bloat the binary when generating a lot of `Decodable` types (apple/swift#60287).

Prior to this change, the default behaviour of CreateAPI is to generated individual enums but we believe that we should generate efficient code by default in this case. 

# Changes

To support this, I've deprecated the old `isGeneratingCustomCodingKeys` option and in it's favour have introduced `isUsingStringsForCodingKeys`. While making this transition, the default behaviour is now flipped meaning that:

1. If you didn't define `isGeneratingCustomCodingKeys` in your config, you will need to set `isUsingStringsForCodingKeys` to `false` to match your previously generated output
2. If you did set `isGeneratingCustomCodingKeys` to `false`, you don't need to do anything apart from remove the flag before it is removed in 0.2.0

Because we're changing the default behaviour, this also results in a lot of changes to the expected tests, hence the big diff 🙏 You can however focus most of the review outside of the **Tests/CreateAPITests/Expected/** directory

## Release Note

- Generated entities will now use string literals via a generated `StringCodingKey` type by default instead of custom `CodingKey` enums. To match behaviour prior to 0.1.0, set `entities.isUsingStringsForCodingKeys` to `false`. For more information, see [#75](https://github.com/CreateAPI/CreateAPI/issues/75).

### DIFF
--- a/Tests/CreateAPITests/GenerateOptionsTests.swift
+++ b/Tests/CreateAPITests/GenerateOptionsTests.swift
@@ -605,7 +605,7 @@ final class GenerateOptionsTests: GenerateBaseTests {
         try compare(package: "edgecases-int32-int64")
     }
     
-    func testEdgecasesGenerateCodingKeys() throws {
+    func testEdgecasesGenerateEnumCodingKeys() throws {
         // GIVEN
         let command = try Generate.parse([
             pathForSpec(named: "edgecases", ext: "yaml"),
@@ -613,7 +613,7 @@ final class GenerateOptionsTests: GenerateBaseTests {
             "--package", "edgecases-coding-keys",
             "--config", config("""
             entities:
-                isGeneratingCustomCodingKeys: false
+                isUsingStringsForCodingKeys: false
             """, ext: "yaml")
         ])
                 
--- a/Docs/ConfigOptions.md
+++ b/Docs/ConfigOptions.md
@@ -79,7 +79,7 @@ Below you can find the complete documentation for all available options.
   - [isSkippingRedundantProtocols](#entitiesisskippingredundantprotocols)
   - [isGeneratingInitializers](#entitiesisgeneratinginitializers)
   - [isSortingPropertiesAlphabetically](#entitiesissortingpropertiesalphabetically)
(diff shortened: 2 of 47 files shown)


Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -605,15 +605,15 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+    func testEdgecasesGenerateEnumCodingKeys() throws {
         // GIVEN
         let command = try Generate.parse([
             pathForSpec(named: "edgecases", ext: "yaml"),
             "--output", temp.url.path,
             "--package", "edgecases-coding-keys",
             "--config", config("""
             entities:
-                isGeneratingCustomCodingKeys: false
+                isUsingStringsForCodingKeys: false
```
liamnichols, line 616: Since this test was previously the only one to check the `StringCodingKey` usage, I've now flipped that so that it tests the enum `CodingKey` usage

liamnichols:
Will close this for now. I'm going to approach in two steps:

1. Mass rename properties (#89) 
2. Change default behaviours

Issue #74, opened 2022-08-02
Make `--split` (`-s`) the default behaviour


liamnichols:
# Background 

- #57 

# Description

Bundling everything into just **Paths.swift** and **Entities.swift** prevents the compiler from parallelising build tasks and results in bad compile times for larger projects. 

Therefore bundling the source has little advantage compared to splitting into individual files (`--split`) so we should make this the default. 

We're not ready to remove the option completely, likely we'll replace it with something like `--no-split` or `--merge` but we should put less emphasis on it. 

The one time that I've found it useful is for predicting the output sources so that we can make efficient `BuildToolPlugin`'s. But in the long run, I want to explore ways to break out the evaluation from the generation so that we could compute this ahead of time... Maybe..

liamnichols:
There are two approaches that we could take here:

## Add inversion to `split`

This would keep the `-s`/`--split` flag but also introduce an inversion called `--no-split`. 

If you ran the generator without specifying `-s`/`--split` then it would result in an error:

```
$ create-api generate schema.json
Error: Missing one of: '--split', '--no-split'
Help:  --split <split>  Split output into separate files
```

Pros: Forces the users who didn't previously split to make a conscious decisions on the behaviour that they want
Cons: One extra flag that needs to be provided

## Deprecate `split` and introduce `mergeSources`

For people using `--split`, we'd log a warning that the property was deprecated (and now the default behaviour), but for those not using `--split`, it will result in a change in generator behaviour. 

Pros: Makes splitting the default behaviour
Cons: Breaking change

---

I think that I'm leaning towards a two-step change:

1. Add the `--split` inversion to force people to make a conscious decision in 0.1.0
2. Deprecate `--split` and `--no-split` in favour of `--merge-sources` in a future release

The result here is that there are no unexpected surprises if you don't read the release notes. If you want to continue merging your sources, you see an error, append `--no-split` and continue about your day. If you were already using `--split`, it makes no difference. 

Then in a future update, we introduce `--merge-sources` and deprecate `--split`/`--no-split`, log a warning for a release or two and then remove the options.

liamnichols:
Removing the 0.1 milestone since this will be a longer-term project. The first phase however is going into 0.1 (see PR above)

LePips:
[...]

PR #87, opened 2022-08-04
[BREAKING] Require `--split` or `--no-split` flag

liamnichols:
# Background 

- #74 
- #57 

To help guide people to generate efficient code, we want to make splitting source code into multiple files the default behaviour. This however would be a breaking change that would bring with it potential confusion if made on its own.

Instead, we should consider making the decision to split sources a conscious one during a transitional phase.

# Changes

In this change, I update the `split` flag to remove the default value of `false` and instead define an inversion that introduces the `--no-split` flag.

This means that when running `create-api` without specifying `-s`, `--split` or `--no-split`, you will now get an error:

```
$ create-api generate schema.json 
Error: Missing one of: '--split', '--no-split'
Help:  --split <split>  Split output into separate source files.
Usage: create-api generate [<options>] <input> --split
  See 'create-api generate --help' for more information.
```

If you already used `-s` or `--split` everything is ok, but if you didn't, you will be forced to make a decision:

1. Continue with `--no-split`
2. Transition to using `--split`

To help guide users about this decision, I added some more discussion to the flag when using `--help`:

```
$ create-api generate schema.json --help
USAGE: create-api generate [<options>] <input> --split

ARGUMENTS:
  <input>                 The OpenAPI spec input file in either JSON or YAML format

OPTIONS:
  --output <output>       The output folder (default: ./.create-api/)
  --config <config>       The path to generator configuration. If not present, the
                          command will look for .create-api.yaml in the current
                          directory. (default: ./.create-api.yaml)
  -s, --split/--no-split  Split output into separate source files.

        Merging the source files offers a compact output, but prevents the compiler
        from parallelizing build tasks resulting in slower builds for larger
        schemas.

        It's recommended that you use --split, but behaviour prior to 0.1.0 matched
        --no-split.

        In a future version, these flags will be deprecated and replaced with an
        alternative opt-out flag instead. For more information, see
        https://github.com/CreateAPI/CreateAPI/issues/74

  -v, --verbose           Print additional logging information
  ...
```

As a result in 0.1.0 you are forced to make a decision. In a future version however, we will relax this again by deprecating `--split` and `--no-split` and replacing it with a more clearly named `--merge-sources` flag which defaults to `false`. 

## Release Note

- You are now required to explicitly pass the `--split` or `--no-split` flag when running the generator. `--no-split` matches the default behaviour prior to 0.1.0 but it is recommended to split your sources for improved build times of your generated code. For more information, see [#74](https://github.com/CreateAPI/CreateAPI/issues/74).

### DIFF
--- a/Sources/CreateAPI/Generate.swift
+++ b/Sources/CreateAPI/Generate.swift
@@ -20,8 +20,24 @@ struct Generate: ParsableCommand {
     @Option(help: "The path to generator configuration. If not present, the command will look for .create-api.yaml in the current directory.")
     var config = "./.create-api.yaml"
 
-    @Flag(name: .shortAndLong, help: "Split output into separate files")
-    var split = false
+    @Flag(
+        name: .shortAndLong,
+        inversion: .prefixedNo,
+        help: ArgumentHelp("Split output into separate source files.", discussion: """
+
+        Merging the source files offers a compact output, but prevents the compiler \
+        from parallelizing build tasks resulting in slower builds for larger schemas.
+
+        It's recommended that you use --split, but behaviour prior to 0.1.0 matched \
+        --no-split.
+
+        In a future version, these flags will be deprecated and replaced with an \
+        alternative opt-out flag instead. For more information, see \
+        https://github.com/CreateAPI/CreateAPI/issues/74
+
+        """)
+    )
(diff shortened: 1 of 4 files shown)


liamnichols:
Closing in favour of #95

Issue #60, opened 2022-07-26
Split filename-template for both entities and paths


LePips:
Using the below files, everything is freely available:
- OpenAPI spec: https://api.jellyfin.org/openapi/jellyfin-openapi-stable.json
- CreateAPI config file: https://pastebin.com/mkYS7wKQ

Command used:
```
create-api generate --output . --config createapi-config.yaml --package JellyfinAPI -s jellyfin-openapi-stable.json"
```

With the `operations` paths style I get conflicting filenames between entities and paths. I actually need to use the `operations` paths style because `rest` results in another error because the spec has a path literally called "Path" and that creates a different issue. You can simply change the paths style and generate to see the error.

Anyways, the only way to properly avoid these file conflicts is to use `--entityname-template "$0Model"` which properly renames all entity files and the corresponding entities, however I would not like to have every entity to be that same name due to convenience. I think a flag change for the `--filename-template <filename-template>` would be more appropriate to split into two flags:
- `--entity-filename-template <entity-filename-template>`
- `--paths-filename-template <paths-filename-template>`

These flags would create a further separation between the two types and aid in these types of conflicts. If this gets greenlit I'll be more than happy to implement to get comfortable with this project.

liamnichols:
Thanks for raising this @LePips! 

> Anyways, the only way to properly avoid these file conflicts is to use `--entityname-template "$0Model"` which properly renames all entity files and the corresponding entities, however I would not like to have every entity to be that same name due to convenience

Yeah I agree. It's not common to suffix model files with **FooModel.swift**. It would make much more sense to add the suffix to the paths (i.e **GetFooRequest.swift** or similar).

> I think a flag change for the `--filename-template <filename-template>` would be more appropriate to split into two flags

Generally this seems like a reasonable idea, but the filename template is also used on some smaller extension files that occasionally get generated too, so not everything falls into the 'paths' or 'entities' bucket.  

In general, I did want to move the `--filename-template` (amongst other options) into the config file rather than command line arguments (#52) so I wonder if there would be a better way to model more specific options in the config file itself? But without thinking a bit more, I don't really have an answer to that yet. 

As an alternative workaround, had you considered trying to rename the conflicting operations in the configuration file? Does that work how I'm thinking it would? 

```yaml
rename:
  operations:
    myConflictingOperationName: someOtherName
```

LePips:
> flag vs config

I'm all for having everything in the config.

> smaller extension files

[...]

Issue #47, opened 2022-06-29
All command line arguments that alter output should be configurable in config file


liamnichols:
Currently, we have a very vast amount of configuration options that must be included in the **create-api.yaml** file, but we also have some other options that also alter how the tool generates the output code that must be passed via CLI args. This includes the following:

- `-s, --split` - Split output into separate files
- `--package <package>` - Generates a complete package with a given name
- `--module <module>` - Use the following name as a module name
- `--vendor <vendor>` - Enabled vendor-specific logic (supported values: "github") 
- `--generate <generate>` - Specifies what to generate (default: paths, entities)
- `--filename-template <filename-template>` - Example: "%0.generated.swift" will produce files with the following names: "Paths.generated.swift". (default: %0.swift) 
- `--entityname-template <entityname-template>` - Example: "%0Generated" will produce entities with the following names: "EntityGenerated". (default: %0) 

This means that there are two sources to dig through when altering configuration, which has the potential to cause confusion. It also makes it hard to invoke `create-api` from multiple different places without either copying the desired arguments or wrapping the invocation in a script. 

Firstly, I'd like to understand if there is a good reason that these options should be separate to the options in the configuration file? 

If we agree that isn't the case, I'd like to start making them available in the configuration instead. 

Unless there is a good reason to keep the arguments as overrides to configuration options, my suggestion would be to deprecate them now and remove them in a future release. Printing a warning when used until they are removed.

liamnichols:
I guess that one question could be, how many different schemas are we generating per configuration? 

It might make sense to invoke the command against multiple schemas, changing only the Package name but reusing the same configuration? 

But if this is the case, I might be more inclined to look for a more generic way to accept config file overrides instead? 

If users were doing this, we'd find out during the deprecation window and would be able to decide accordingly then

liamnichols:
I'm moving this up to 0.1.0 as I think it could be an important breaking change to get right. As it stands, I want to move the following:

Argument|Config
---|---
`--split`/`-s`|`isMergingSources`
`--package`|`mode` & `target`
`--module`|`mode` & `target`
`--generate`|`entities.isEnabled` & `paths.isEnabled`
`--filename-template`|`rename.files`
`--entityname-template`|`rename.entities`
`--vendor`|`vendor`

For example, the following invocation:

```bash
[...]

PR #95, opened 2022-08-04
[BREAKING] Replace `--split` with `--merge-sources` and Move generated extensions into their own source files

LePips:
- Overtakes https://github.com/CreateAPI/CreateAPI/pull/87
- Closes https://github.com/CreateAPI/CreateAPI/issues/74
- Part of https://github.com/CreateAPI/CreateAPI/issues/60

## What
- Makes splitting files the default and makes a new `--merge-sources` flag
- Creates separate extension files per purpose at the top level instead of merging them
- Updates tests to reflect


liamnichols (commented):
Great work on this, thanks for picking it up @LePips! 

What is the notification however for moving the extensions out of **Entities.swift** and **Paths.swift** when **using `--merge-sources`**? It makes sense when splitting, but this one looks a little odd:

Screenshot 2022-08-04 at 20 42 28

I think I would have expected these to stay inside the merged file?

---

Secondly, I just want to bring this to your attention: https://github.com/CreateAPI/CreateAPI/issues/47

I'm thinking that maybe we can just go all of the way and move `--merge-sources` into the configuration options instead? WDYT about that approach? 

There is also another part of that ticket that is relevant to you, which is about how to go about describing the filename templates in the config too so any feedback is welcomed 👍

LePips:
> extension using `--merge-sources`

I was under the impression that they should be broken out no matter what from https://github.com/CreateAPI/CreateAPI/issues/60 and personally think they should be. The only use case I can think of for merging schema-generated objects/paths would be for reference. While devs might also move files around, they will still have to be responsible for moving these extensions as well (split or merged).

Essentially, it splits the concerns of each file: `Entities.swift` for generated objects, `Paths.swift` for generated paths, and then each extension file is their own helper file.

> #47

I don't think that this simple change in logic should worry about that until a consensus is reached there. I'll leave my thoughts on the issue.

liamnichols:
> I was under the impression that they should be broken out no matter wha

Doh, you are right.. I forgot that I even suggested that 😅  https://github.com/CreateAPI/CreateAPI/issues/60#issuecomment-1203141626 

> I don't think that this simple change in logic should worry about that until a consensus is reached there. I'll leave my thoughts on the issue.

Sure 👍 

---

Let me just do some final tests of this change 👀 It's hard to review on GitHub as the diff freezes up my browser 😄

Sources/CreateAPI/Generate.swift
```diff
@@ -20,8 +20,11 @@ struct Generate: ParsableCommand {
     @Option(help: "The path to generator configuration. If not present, the command will look for .create-api.yaml in the current directory.")
     var config = "./.create-api.yaml"
 
-    @Flag(name: .shortAndLong, help: "Split output into separate files")
-    var split = false
+    @Flag(help: ArgumentHelp("Merge Entities and Paths into single output files", discussion: """
+                            Merging the source files offers a compact output, but prevents the compiler \
+                            from parallelizing build tasks resulting in slower builds for larger schemas.
+                            """))
```
liamnichols, line 26: Should only be one level of indentation here

```suggestion
    @Flag(help: ArgumentHelp("Merge Entities and Paths into single output files", discussion: """
        Merging the source files offers a compact output, but prevents the compiler \
        from parallelizing build tasks resulting in slower builds for larger schemas.
        """))
```

liamnichols (commented):
It looks like the old **Tests/CreateAPITests/Expected/petstore-split** directory is still here, can we delete it please 🙏

liamnichols:
~Please can you revert the rest of the changes to **Snapshots.swift** 🙏~ Keep in mind that this diff is gigantic and barely loads in the browser so it makes it hard to efficiently review things 🙇

Its not worth changing it now, but my suggestion if we were to do this again would be to submit a pull request that switches `--split` for `--merge-sources`, but updates all the test fixtures that didn't use `--split` to use `--merge-sources` (and vice versa).. This would have helped us to be completely sure that the only other changes in output were related to moving the extensions since the diff would have been readable. We could then update all the fixtures to stop using `--merge-sources` in a separate PR where the diff is less important 🙏 

Thanks again though for working on this! Will merge it once the checks finish ✅

liamnichols:
Thanks again @LePips for the great contribution 🚀

Issue #98, opened 2022-08-04
Simplify `comments` configuration


liamnichols:
The `comments` configuration is currently an object with 5 customisations expressed as Bool and 1 main `isEnabled` flag.

In my head, we could express this the same way with an enum/option set of values and a single `commentFeatures` array/set.. Something like the following:

**Disable Comments:**
```yaml
commentFeatures: []
```

**Everything:**
```yaml
commentFeatures:
- title
- description
- example
- externalDocumentation
- capitalized
```

**Description (capitalized):**
```yaml
commentFeatures:
- description
- capitalized
```

In Swift, this would be something like the following:

```swift
public enum CommentFeature: String {
    case title, description, example, externalDocumentation, capitalized
}

public var commentFeatures: Set<CommentFeature> = [.title, .description, .example, .externalDocumentation, .capitalized]
```

Alternatively I'll just deswiftify the boolean properties on the existing object.

Issue #97, opened 2022-08-04
Simplify `acronym` related properties


liamnichols:
Currently there are four things related to acronyms:

1. The default list (`["url", "id", "html", "ssl", "tls", "https", "http", "dns", "ftp", "api", "uuid", "json"]`)
2. `isReplacingCommonAcronyms` (defaults to `true`)
3. `addedAcronyms`
4. `ignoredAcronyms`

Its done like this to include a default set of acronyms, and then allow customisation by either disabling replacement, ignoring specific or adding new ones.

After thinking it through some more, this could be simplified down to a single configuration option:

```swift
/// A set of acronyms that should be uppercased if detected.
///
/// If you don't want to uppercase any acronyms, set this property to an empty array.
public var uppercaseAcronyms: [String] = ["url", "id", "html", "ssl", "tls", "https", "http", "dns", "ftp", "api", "uuid", "json"]
```

We provide (and document) the default set, so if the user wants to add to it, they can incorporate the set into their configuration file. 

If the user doesn't want to uppercase acronyms, they can set the property to an empty array. 

I can't see any reason not to do this?

LePips:
Yeah that definitely cleans things up

Issue #94, opened 2022-08-04
Remove option `isSwiftLintDisabled`?


liamnichols:
While it could sometimes be considered a useful feature, I have to slight concerns with it:

1. It assumes we use SwiftLint, but what about other tools?
2. It's more efficient to use a **.swiftlint.yml** configuration to setup `ignore` rules for the generated output directory.
3. If we want it, we could just add it to `fileHeaderComment` instead.

Any other reasons to keep it? I'm suggesting that we either:

1. Remove `isSwiftLintDisabled` with no replacement
2. Remove `isSwiftLintDisabled` but add `// swiftlint:disable all` into the `fileHeaderComment`

Keen to hear what others think.

LePips:
I personally thought this was weird as well. With the discussion from this thread, https://github.com/CreateAPI/CreateAPI/pull/82#discussion_r936799801, we should leave it up to the developer to do whatever linting since they will probably edit it themselves to do post-processing. I personally will be with SwiftFormat.

Issue #93, opened 2022-08-04
Remove option `isAdditionalPropertiesOnByDefault`?


liamnichols:
When renaming some config options, I noticed that `isAdditionalPropertiesOnByDefault` seems to be redundant. 

It is currently used like so:

https://github.com/CreateAPI/CreateAPI/blob/e0cb7dbb11ced91c54c10291ce992acd3fc49e46/Sources/CreateAPI/Generator/Generator%2BSchemas.swift#L400-L406

So when a schema does not define `additionalProperties`, if there are no other properties in the schema and `isAdditionalPropertiesOnByDefault ` is set to `true` (the default), it will assume that `additionalProperties` is to be treated as `true`.

According to the OpenAPI Spec however, `additionalProperties` has a default value of `true` anyway:

> - `additionalProperties` - Value can be boolean or object. Inline or referenced schema MUST be of a [Schema Object](https://swagger.io/specification/#schema-object) and not a standard JSON Schema. **Consistent with JSON Schema, `additionalProperties` defaults to `true`.**

https://swagger.io/specification/

This means that we shouldn't need to assume a missing `additionalProperties` (nil) is ambiguous because according to the specification it should be treated as `true`. 

If I am not mistaken, we can remove the option and simplify the implementation to the following:

```diff
- var additional = details.additionalProperties 
- if details.properties.isEmpty, options.entities.isAdditionalPropertiesOnByDefault { 
-     additional = additional ?? .a(true) 
- } 
- guard let additional = additional else { 
-     return nil 
- }
+ let additional = details.additionalProperties ?? .a(true) 
```

The only thing I am not sure of is why we check `details.properties.isEmpty`. I removed the code and all tests pass so as far as I can tell, this change has no impact.

PR #99, opened 2022-08-04
Rename all swifty-style boolean configuration properties

liamnichols:
# Background

- Part Of #89 

Property names such as `isDoingFoo` are out of place in YAML configuration files so for 0.1 we plan to normalise them all.

# Changes

In this change, I rename the following properties:

Before|After
---|---
`isUsingIntegersWithPredefinedCapacity`|`useIntegersWithPredefinedCapacity`
`isNaiveDateEnabled`|`useNaiveDate`
`isPluralizationEnabled`|`pluralizeProperties`
`isInliningTypealiases`|`inlineTypealiases`
`isGeneratingSwiftyBooleanPropertyNames`|`useSwiftyPropertyNames`
`isGeneratingEnums`|`generateEnums`
`isAddingDeprecations`|`annotateDeprecations`
`isStrippingParentNameInNestedObjects`|`stripParentNameInNestedObjects`
`isInliningPropertiesFromReferencedSchemas`|`inlineReferencedSchemas`
`isAddingDefaultValues`|`defaultValues`
`isSortingPropertiesAlphabetically`|`sortPropertiesAlphabetically`
`isGeneratingEncodeWithEncoder`|`alwaysIncludeEncodableImplementation`
`isGeneratingInitWithDecoder`|`alwaysIncludeDecodableImplementation`
`isGeneratingInitializers`|`includeInitializer`
`isSkippingRedundantProtocols`|`skipRedundantProtocols`
`isGeneratingIdentifiableConformance`|`identifiableConformance`
`isGeneratingMutableStructProperties`|`mutableStructProperties`
`isGeneratingMutableClassProperties`|`mutableClassProperties`
`isMakingClassesFinal`|`makeClassesFinal`
`isGeneratingStructs`|`generateStructs`
`isRemovingRedundantPaths`|`removeRedundantPaths`
`isMakingOptionalPatchParametersDoubleOptional`|`makeOptionalPatchParametersDoubleOptional`
`isInliningSimpleQueryParameters`|`inlineSimpleQueryParameters`
`isInliningSimpleRequests`|`inlineSimpleRequests`
`isGeneratingResponseHeaders`|`generateResponseHeaders`

And I replace the following:

Before|After
---|---
`isGeneratingCustomCodingKeys`|`optimizeCodingKeys`

I left a few that I will followup on individually:

- #92 
- #93 
- #94 
- #97 
- #98


kean:
Hey, @liamnichols. I reviewed the new names and sharing a few thoughts. I don't have any strong feelings about these names though.

- `useIntegersWithPredefinedCapacity` - these are called "fixed-width integers'
- `stripParentNameInNestedObjects` - this is probably safe to remove; it has mechanisms to ensure that it never (?) leads to ambiguity
- `defaultValues` - should it be a verb, e.g. add/use default values
- `includeInitializer` - I'd probably use "generate" for this, or "addInitializers" but either way is good
- `identifiableConformance` - same as the previous
- `generateStructs` - I don't remember what it does anymore, but it seems like an enum would've probably been a better option
- `makeOptionalPatchParametersDoubleOptional` - that's a mouthful, but it's not yet in its target state – using double-optionals is not ideal
- `generateResponseHeaders` - seems good, but inconsistent with other options, e.g. `includeInitializer`; they should probably use the same verb – "generate" sounds good (but not in "generateStructs")

liamnichols:
Thanks @kean for the feedback! 

> - `useIntegersWithPredefinedCapacity` - these are called "fixed-width integers'

This is good to know, in that case I think `useFixWidthIntegers` seems more appropriate. 

> - `stripParentNameInNestedObjects` - this is probably safe to remove; it has mechanisms to ensure that it never (?) leads to ambiguity

Ok thanks, I'll take a look at it but will probably leave it for now as I'm still not too sure what it does exactly 😄 

> - `defaultValues` - should it be a verb, e.g. add/use default values?

Good point, maybe to align with the others `includeDefaultValues`

> - `includeInitializer` - I'd probably use "generate" for this, or "addInitializers" but either way is good

heh yeah I was a little stuck on this. To be honest, I was trying not to use the term `generate` in any option but looking back it, I was a little inconsistent still 😄 

I feel like generate is a bit repetitive because these are options for the generator hence going for a term like include instead. "Include the initialisers when generating entities" for example. 

> - `identifiableConformance` - same as the previous

Yeah this too could be `includeIdentifiableConformance`, I'll change it. 

> - `generateStructs` - I don't remember what it does anymore, but it seems like an enum would've probably been a better option. Otherwise it's not clear. For example, what if it's false? Will it just skip and not generate any structs?

In #100 we decided to do like you suggested and replace it with an `type` enum that is a bit more descriptive when it's false (`type: struct`, `type: class` or `type: finalClass`)

> - `makeOptionalPatchParametersDoubleOptional` - that's a mouthful, but it's not yet in its target state – using double-optionals is not ideal

Heh yeah. I've actually left this one undocumented in ConfigOptions.md because I want us to revisit it (part of #18 maybe). I think its fine how it is for 0.1 but it will likely change again soon.

> `generateResponseHeaders` - seems good, but inconsistent with other options, e.g. includeInitializer; they should probably use the same verb – "generate" sounds good (but not in "generateStructs")

I think also `includeResponseHeaders` might make sense here (see my reasoning for 'include' vs 'generate' above)

PR #105, opened 2022-08-04
Remove `isAdditionalPropertiesOnByDefault` 

### LINKED ISSUE #93: Remove option `isAdditionalPropertiesOnByDefault`?

liamnichols:
- Closes #93 

If I understand the schema correctly, we don't need to worry about `additionalProperties` meaning anything other than `true` when it's not specified because this is documented in the OpenAPI spec:

> - `additionalProperties` - Value can be boolean or object. Inline or referenced schema MUST be of a [Schema Object](https://swagger.io/specification/#schema-object) and not a standard JSON Schema. **Consistent with JSON Schema, `additionalProperties` defaults to `true`.**

This would mean that `isAdditionalPropertiesOnByDefault` is redundant. Therefore I am removing it and making it's behaviour when `true` the default.

### DIFF
--- a/Docs/ConfigOptions.md
+++ b/Docs/ConfigOptions.md
@@ -84,7 +84,6 @@ Below you can find the complete documentation for all available options.
   - [optimizeCodingKeys](#entitiesoptimizecodingkeys)
   - [defaultValues](#entitiesdefaultvalues)
   - [inlineReferencedSchemas](#entitiesinlinereferencedschemas)
-  - [isAdditionalPropertiesOnByDefault](#entitiesisadditionalpropertiesonbydefault)
   - [stripParentNameInNestedObjects](#entitiesstripparentnameinnestedobjects)
   - [exclude](#entitiesexclude)
   - [include](#entitiesinclude)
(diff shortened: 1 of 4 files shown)


kean:
I haven't encountered any specs where I needed it. Not sure what was the context behind adding it. I think it's safe to remove.

Issue #71, opened 2022-08-01
Exclude for Individual Properties


LePips:
I have a very large object definition that contains itself by reference on a single property. The generator will currently turn this object into a `class` when I need a `struct` for usage in my app (just strongly prefer value types in my case). In my specific case I can just ignore the property that is causing this issue entirely, as I have done previously with editing the config file.

With the following example, I can just ignore `best_friend`:
```yaml
    Pet:
      title: A pet title
      description: A pet description
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          example: "Buddy"
          type: string
        tag:
          type: string
        best_friend:
          $ref: "#/components/schemas/Pet"
          description: "Best friend of your pet!"
```

```yaml
entities:
  - excludeProperties: ["Pet.best_friend"]
```

Will create an expected `struct` with no `bestFriend` property.

This is more of a discussion if `excludeProperties` is a good config name and whether the analog `includeProperties` should be created. In both cases, I know that I would have to check against the existing `exclude/include` lists to see if the object should be included in the overall generation. I think an error should be thrown in cases where usage of these flags together don't make sense. Or, we can just ignore these and leave it up to the user to have a clean config file.

```yaml
entities:
  # file 1
  - include: ["some-other-object"]
  - excludeProperties: ["Pet.best_friend"] # Error as "Pet" is not included in generation
  
  # file 2
  - include: ["some-other-object"]
  - includeProperties: ["Pet.best_friend"] # Error as "Pet" is not included in generation
  
  # file 3
  - exclude: ["Pet"]
  - excludeProperties: ["Pet.best_friend"] # Error as "Pet" is not included in generation
  
  # file 4
  - exclude: ["Pet"]
  - includeProperties: ["Pet.best_friend"] # Error as "Pet" not included in generation
```

Or instead of a new flag we change the sections to:
```yaml
entities:
  exclude:
    - objects: [...]
    - properties: [...]
  include:
    - objects: [...]
    - properties: [...]
```

### Refactor Idea

A brainstormed refactor to `include/exclude` could be the following:
```yaml
entities:
  # file 1
  - include:
    - Pet: # generates the "Pet" object with only the following properties
      - properties:
        - id
        - name
    - Store: # generates the "Store" object with no properties as it is empty
      - properties: 
    - Error: ... # need some way to indicate "Error" object generation with all properties included?

  # file 2
  - exclude:
[...]

PR #106, opened 2022-08-05
Exclude Individual Entity Properties

LePips:
- Closes https://github.com/CreateAPI/CreateAPI/issues/71

Adds basic property exclusions for entities. I really don't know why I overthought this as it's really simple. I provide a warning for when an excluded property in the config does not actually exist in the schema as it's technically incorrect against the schema, but won't break the generation. I don't provide any other warnings or errors for exclusion as that will be caught either at the compilation or the API level by the developer _and_ they manually excluded objects in their config.

I don't provide a warning against renamings+excludes or redundant excludes, which would require validating the exclude list right after parsing and before any generation. I think these are up to the developer to keep clean or we can create a robust config validator to surface these warnings.


Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -337,7 +337,16 @@ extension Generator {
...
     private func makeInlineProperties(for type: TypeName, object: JSONSchema.ObjectContext, context: Context) throws -> [Property] {
-        var keys = object.properties.keys
+        let excludedProperties = excludedProperties(for: type.rawValue)
+        let excludedDiff = excludedProperties.subtracting(object.properties.keys)
+        
+        if !excludedDiff.isEmpty {
+            for diff in excludedDiff {
+                try handle(warning: "Excluded property \(type.rawValue).\(diff) does not exist on schema object \(type.rawValue)")
+            }
+        }
```
liamnichols, line 347: I know we do it elsewhere, but does calling `for diff in excludedDiff { ... }` when the array is empty actually take a hit in terms of performance? 

I don't know the answer myself but it feels like it would be a no 😄

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -349,6 +358,14 @@ extension Generator {
             }
         }
     }
+    
+    private func excludedProperties(for type: String) -> Set<String> {
+        let excluded = options.entities.exclude
+            .map { $0.split(separator: ".") }
+            .filter { $0.count == 2 && $0[0].elementsEqual(type) }
+            .map { String($0[1]) }
+        return Set(excluded)
```
liamnichols, line 367: So this got me thinking a little about performance, but we don't really have any benchmarking in place and I can't see a significant problem here although I was trying to think of ways to pre-process this as much as possible when deserialising the config. 

I then also remembered the current implementation for entity name templating:

https://github.com/CreateAPI/CreateAPI/blob/83a871b63126ebf1978f28a13a026bdac2a4a2e9/Sources/CreateAPI/Generate.swift#L146-L149

As it stands, this would break because it assumes that the elements inside `exclude` are specific entity names and doesn't expect the dot notation for properties too. 

With that, I had an idea that maybe we can use a `RawRepresentatble` backed type instead of a string for this config option:

```swift 
public struct EntityExclude: RawRepresentable, Codable, Hashable {
    public var name: String
    public var property: String?

    public init(name: String, property: String?) {
        self.name = name
        self.property = property
    }

    public init?(rawValue: String) {
        let components = rawValue.components(separatedBy: ".")
        switch components.count {
        case 1:
            self.init(name: components[0], property: nil)
        case 2:
            self.init(name: components[0], property: components[1])
        default:
            return nil
        }
    }

    public var rawValue: String {
        if let property = property {
            return "\(name).\(property)"
        } else {
            return name
        }
    }
}
```

```swift
public var exclude: Set<EntityExclude> = []
```

With that, you should be able to property apply the entity name template and it also saves a little processing in this lookup here.

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -657,4 +657,31 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+            "--config", config("""
+            entities:
+                exclude:
+                - Error
+                - Pet.id
+                - Store.pets
+            rename:
+                properties:
+                    Pet.id: notID
+                    Pet.name: id
```
liamnichols, line 677: So if I understand this right, the logic has ended up excluding `notID` and included `id`. I guess that makes sense since its consistent with how entity naming templates work. 

For the sake of the test example, can you also please use `--entityname-template "%0DTO"` or something similar to confirm its working as expected when entities are renamed too? 🙇

liamnichols (commented):
Great job! Yeah it did turn out pretty simple in the end 👍

Sources/CreateAPI/Generator/Generator+Schemas.swift
LePips: Nope, in fact to be pedantic it adds ~150 instructions for the case where it isn't empty.

Tests/CreateAPITests/GenerateOptionsTests.swift
LePips: No, as the excluded filter happens first and then renaming happens in the next step when the property is made in `makeProperty`. So actually, the _original_ `Pet.id` was removed and then `Pet.name` is renamed to _id_. This was proven with iterative testing and would have been better shown with different property types (both are strings).

I personally think that these are the correct steps. `exclude` should only worry about excluding objects directly from the schema and the rename happens for entity generation. This is further backed up by the fact that we require renames to be in the format of the generated entity name.

Sources/CreateAPI/Generator/Generator+Schemas.swift
LePips: Good catch and I approve of the `RawRepresentable` for pre-processing. I can take a look at this.

Tests/CreateAPITests/GenerateOptionsTests.swift
LePips: > please use `--entityname-template "%0DTO"` ... to confirm

Good catch as this does not work. The entity template is applied as jobs are created to generate the entity so the object type name is different than the exclude. However, I don't think that this should be correct. I think that the `entityname-template` should be applied _last_ as the entity is created for a similar reason as above for `exclude` and `rename`: use the schema object name for setting the excludes and renames. 

I don't think it is correct to specify:

```yaml
# using --entityname-template "$0Model.swift"

 entities:
   exclude:
   - PetModel.id
```
Additionally, I think `--entityname-template` [should be removed](https://github.com/CreateAPI/CreateAPI/issues/47#issuecomment-1205703517).

Sourcery/ConfigOptions.md.stencil
```diff
@@ -41,7 +41,7 @@
 ## {{ type.localName|lowercase }}.{{ variable.name }}
 {% endif %}
 
-**Type:** {{ variable.typeName|replace:"?","" }}<br />
+**Type:** {{ variable.typeName|replace:"?",""|replace:"EntityExclude","String" }}<br />
```
liamnichols, line 44: I wouldn't worry about doing this, as long as the documentation is good and has examples. 

If we want to make the 'type' accurate for yaml/json then we can do it universally

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -99,7 +99,8 @@ extension Generator {
             return options.entities.include.contains(name)
         }
         if !options.entities.exclude.isEmpty {
-            return !options.entities.exclude.contains(name)
+//            return !options.entities.exclude.contains(name)
+            return true
```
liamnichols: Debugging?

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -337,7 +337,16 @@ extension Generator {
     }
 
     private func makeInlineProperties(for type: TypeName, object: JSONSchema.ObjectContext, context: Context) throws -> [Property] {
-        var keys = object.properties.keys
+        let excludedProperties = options.entities.exclude
+            .filter { $0.name == type.rawValue }
+            .compactMap { $0.property }
+        let excludedDiff = Set(excludedProperties).subtracting(object.properties.keys)
```
liamnichols, line 343:
```suggestion
        let unknownProperties = Set(excludedProperties).subtracting(object.properties.keys)
```

or something that's a bit of descriptive maybe?

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -337,7 +337,16 @@ extension Generator {
...
+        
+        for diff in excludedDiff {
+            try handle(warning: "Excluded property \(type.rawValue).\(diff) does not exist on schema object \(type.rawValue)")
```
liamnichols, line 346: I'm just wondering if this could be a bit more explanatory? 

```diff
- Excluded property User.id does not exist on schema object User
+ Invalid entity exclude 'User.id'. Property 'id' does not exist on schema 'User'
```

```suggestion
            try handle(warning: "Invalid entity exclude '\(type.rawValue).\(diff)'. Property '\(diff) does not exist on schema '\(type.rawValue)'")
```

Sources/CreateAPI/Generator/Generator+Schemas.swift
LePips: Oops

LePips: Your review seems to be attached to a certain commit that I did while working on this. This is not in the final version.

Sourcery/ConfigOptions.md.stencil
LePips: I strongly recommend we only present only yaml acceptable values. `EntityExclude` is entirely for our development purposes and won't make sense to the user reading documentation even with examples.

LePips: Reading up on Sourcery annotations. I might be able to make a type replacement annotation.

LePips: I'll drop this point for the implementation of this feature. However, I still propose we make surfacing documentation types their yaml types instead of our underlying types.

liamnichols: We could add a custom `'defaultValue'` annotation to use instead but it might not be worth worrying about this too much right now, I'm trying out yet another option for handling documentation although I'm not too sure if it will work just yet or not 😅

liamnichols (approved):
Another great contribution, thanks @LePips!

PR #108, opened 2022-08-05
Split Filename Template

LePips:
- Closes https://github.com/CreateAPI/CreateAPI/issues/60

Entities and paths now simply have their own `filenameTemplate` config option. This does not affect extension files. The original `--filename-template` flag was removed.


Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -657,4 +641,26 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
         try compare(package: "petstore-identifiable")
     }
+    
+    func testPetstoreFilenameTemplate() throws {
+        // GIVEN
+        let command = try Generate.parse([
+            pathForSpec(named: "petstore", ext: "yaml"),
+            "--output", temp.url.path,
+            "--package", "petstore-filename-template",
+            "--generate", "entities",
```
liamnichols, line 651: Can we test generating both?

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -18,22 +18,6 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
-            "--output", temp.url.path,
-            "--package", "petstore-change-filename",
-            "--filename-template", "%0.generated.swift"
-        ])
-        
-        // WHEN
-        try command.run()
-        
-        // THEN
-        try compare(package: "petstore-change-filename")
```
liamnichols, line 34: Please be sure to remove **Expected/petstore-change-filename** as well as the dependency from in **AllPackages/Package.swift**

LePips: Is there a way we can automate the population of that file?

Tests/CreateAPITests/GenerateOptionsTests.swift
LePips: Whoops, definitely should.

Tests/CreateAPITests/GenerateOptionsTests.swift
liamnichols: In theory I think so, but its a little tricky so I don't have an exact answer as to how though

PR #113, opened 2022-08-05
Introduce replacement entity type configurations

### LINKED ISSUE #100: Simplify Entity-type and Mutable Properties in Config
We currently have two config options for entity-type generation, `entities.isGeneratingStructs` and `entities.isMakingClassesFinal`. Instead of these two flags, we condense into an enum:
```yaml
entities:
  type: struct || class || finalClass
```

Still preserve `entities.entitiesGeneratedAsClasses` and `entities.entitiesGeneratedAsStructs` for now.

Additionally, there are two config options for property generation: `entities.isGeneratingMutableClassProperties` and `entities.isGeneratingMutableStructProperties`. Condense into a simple bool:
```yaml
entities:
  mutableProperties: true || false # default true
```

liamnichols:
- Closes #100 

Before this change, to customise generation between structs, classes and classes marked as `final`, you would use the following options:

- `isGeneratingStructs` - to switch between structs and classes as the default
- `isMakingClassesFinal` - to make classes `final` if they were generated 
- `entitiesGeneratedAsClasses` to provide overrides that should always generate as a class
- `entitiesGeneratedAsStructs` to provide overrides that should always generate as a struct

For the 0.1 release, we want to simplify this into two properties:

- `defaultType` - (`[struct, class, finalClass]`) the preferred type used for all entities when generating 
- `typeOverrides` - A dictionary map of entity names and the type that they should be (from the list above)

This reduces overhead in the config file and also gives you more flexibility when choosing between `class` and `final class`.

In addition, we want to simplify the mutable property flags by going from the following:

- `isGeneratingMutableClassProperties` (true or false)
- `isGeneratingMutableStructProperties` (true or false)

to

- `mutableProperties` (`[]`, `[structs, classes]`, `[structs]` or `[classes]`)

In #111 I also updated the config deserialiser so that it is capable of processing `true`, `false` or a single element when working with array/set based options. For example in YAML/JSON, `mutableProperties: true` would be considered the same as `mutableProperties: [structs, classes]`.


liamnichols:
> As a part of this, we should remove the behavior that will implicitly convert a `struct` to a `class` in `Generator+Render.swift - hasReferences()`

What's the motivation? I'm not against the idea, but to me if we can detect this and help keep the code compiling then it sounds preferable to generating code that won't compile?

LePips:
1 - We will be able to exclude the properties that would cause such a problem manually after https://github.com/CreateAPI/CreateAPI/pull/106, if desired
2 - It breaks declaration. I say I want a `struct` but then I get a `class`?

I think it's completely find if it doesn't compile at first as we can set the override manually.

liamnichols:
> 2 - It breaks declaration. I say I want a `struct` but then I get a `class`?

True. In that case I am going to log it as a warning then with instructions on how to resolve it. It feels appropriate since we can easily fallback with the existing logic, and `--strict` mode exists for treating warnings as errors.

LePips:
That's fine as I personally will ignore the problematic properties. The implicit conversion should be documented.

liamnichols:
Done in 7f8c73e90f182f5696d66be9a375da3535064019 👍

PR #118, opened 2022-08-06
`access` as underlying enum

### LINKED ISSUE #66: Make `ConfigOptions.access` an enum
https://github.com/CreateAPI/CreateAPI/blob/baa71afbcfc0c5ba591eab55e07c933357c8a381/Sources/CreateOptions/ConfigOptions.swift#L63-L64

The `access` property is currently just a string meaning that you can put anything in the config file and we'll end up generating non-functioning code. 

CreateAPI will only support `public` or `internal` access levels since `open` can't be applied everywhere and `private` would just generate code that is unusable. We could model this with an enum instead.

One thing to note is that while we could specify `access: internal` in the config, we'd want to generate code without access level specifiers (i.e the same as an empty string) since the `internal` keyword itself in generated code is redundant.

For reference, we've used enums elsewhere before in `ConfigOptions` so the same pattern can be followed:

https://github.com/CreateAPI/CreateAPI/blob/baa71afbcfc0c5ba591eab55e07c933357c8a381/Sources/CreateOptions/ConfigOptions.swift#L88-L97

LePips:
- Closes #66 

Straightforward change. Big change to tests though apparently.


Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -359,6 +359,24 @@ final class GenerateOptionsTests: GenerateBaseTests {
...
+            "--config", config("""
+            access: internal
+            """, ext: "yaml")
+            ])
+            
+        // WHEN
+        try command.run()
+        
+        // THEN
+        try compare(package: "petstore-internal-access-control")
```
liamnichols, line 378: Can we please generate the equivalent of what was in `edgecases-change-access-control`? It makes things easier to review because the diff of the generated package won't have changed. 

Also, before recording the snapshots. Please can you delete the contents of the Expected folder (not the folder itself) so that we cleanup unrelated files. For example, looking at this branch the following shouldn't be there:

Screenshot 2022-08-06 at 13 00 02

Tests/CreateAPITests/AllPackages/Package.swift
```diff
@@ -17,7 +17,6 @@ let package = Package(
     dependencies: [
         .package(path: "../Expected/cookpad"),
         .package(path: "../Expected/discriminator"),
-        .package(path: "../Expected/edgecases-change-access-control"),
```
liamnichols, line 20: Think you also need to include the new package, unless you go back to keeping the same name

liamnichols (commented):
Thanks for picking this up! Just a few suggestions about keeping the tests in shape and making it easier to diff 🙏

Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -65,9 +65,14 @@ import Foundation
 public struct ConfigOptions: Encodable {
     /// The default generator options
     public static let `default` = ConfigOptions()
+    
+    /// Available access controls
+    public enum Access: String, Codable {
+        case `internal`, `public`
+    }
 
     /// Access level modifier for all generated declarations
```
liamnichols, line 74: Lets list the available options here 🙏

Tests/CreateAPITests/GenerateOptionsTests.swift
LePips: By "generate the equivalent" do you mean the following config:
```yaml
access: ""
```

Or do you mean the keep the name of the package the same? The new test is no longer an edge case and follows the other test naming conventions, so the big diff is expected but I just didn't expect that large of a deletion diff from the edge cases package.

liamnichols: Changing the package name is ok since it's just a rename. 

But changing the schema is unnecessary for what we are doing in this PR so it is easier for us to focus on the actual change (adding the enum with zero breaking changes) if we don't generate a lot of extra noise in the diff. 

I know in this particular PR its all pretty low risk, but if I could glance and see from the diff that there are zero changes to the generated output then I can be easily satisfied that the change is good and we will have merged it already. Changing the schema means that there are now loads of deletions and a bunch of new files that I have to actually look at and think about to know for sure if this is doing what we expect. 

It's worth getting into this practice of separating the changes and minimising the diff so that we can get things through quicker with more confidence 🙏

liamnichols: The structure of the snapshot tests need an overall rethink anyway (including practices on easily generating smaller outputs for testing minor details) so I wouldn't worry too much about trying to change things right now

liamnichols (approved):
Thanks @LePips, looks great!

PR #123, opened 2022-08-07
Automatically update AllPackages bundle when regenerating snapshots

liamnichols:
When we recreate the generator snapshots, if we add a new package we usually have to manually update **Tests/CreateAPITests/AllPackages/Package.swift** so that CI will try and build the package when we create Pull Requests. This was a pain to keep on top of. 

In this change, I update the tests so that they record all of the generated packages and upon completion of the tests, they update the **Package.swift** file. 

It's a little rough and ready, but I didn't want to spend a lot of time refactoring this area although I think that we probably could do. Any suggestions for improvements are welcome.

cc @LePips

### DIFF
--- a/Tests/CreateAPITests/Compliling/PackageCompiler.swift
+++ b/Tests/CreateAPITests/Compliling/PackageCompiler.swift
@@ -1,0 +1,67 @@
+import Foundation
+import XCTest
+
+final class PackageCompiler: NSObject {
+    static let shared = PackageCompiler()
+
+    private var dependencies: [PackageManifest.Dependency]
+    private var shouldUpdateAllPackages: Bool
+
+    override init() {
+        self.dependencies = []
+        self.shouldUpdateAllPackages = true
+        super.init()
+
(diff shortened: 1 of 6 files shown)


LePips:
Great stuff! With the inclusion of the `XCTestObservation` conformance do you think we can also delete the `Expected` folder when snapshots are being initially generated?

liamnichols:
> Great stuff! With the inclusion of the `XCTestObservation` conformance do you think we can also delete the `Expected` folder when snapshots are being generated?

I was thinking about it, but was slightly deterred because of the scenario where you might run just one test rather than all of them. But maybe there is a way to detect that.. I'll keep looking at this because after looking again at the code this morning, I think I want to refactor it some more as well 😄

liamnichols:
Took a much cleaner approach now by breaking things out into their own responsibilities. We now have `PackageCompiler` (not sure on that name 😄) to collect the dependencies as we generate the snapshots and then write them out using `PackageManifest` at the end of the tests

PR #126, opened 2022-08-08
Support overriding any generator configuration options using the `--config-option` argument

liamnichols:
- related to #47 

To adjust behaviours for the generated outputs prior to 0.1, you'd have to specify command line arguments for some things or use the configuration file for others. This was not a major issue since the inconsistencies were limited, but moving forward in 0.1 we want every customisation to be sourced from an option that can be defined in the configuration file.

While I don't anticipate it being required in may use cases for CreateAPI. There might still be times where you want to customise just one or two options and therefore don't want the burden of a configuration file. Alternatively, you might need to be able to customise the behaviour of a very specific option when generating sometimes but still want the bulk of the options to be maintained in the configuration file.

To support this, I'm adding support for [ConfigurationParser](https://github.com/liamnichols/swift-configuration-parser)'s `OptionOverride` feature. This lets us parse the configuration file with an array of overrides. In this change, I introduce a `--option` argument that is the `[OptionOverride]` type. These are then passed into the parser and will alter behaviour as desired. 

This can also be a very useful feature to ease a transition as we move the remaining arguments into `ConfigOptions` like we did with `--filename-template`. 

I've also added a test to demonstrate the behaviour.


Sources/CreateAPI/Generate.swift
```diff
@@ -23,6 +24,19 @@ struct Generate: ParsableCommand {
         """))
     var config = ConfigFileLocation()
 
+    @Option(help: ArgumentHelp("Options to be applied when generating.", discussion: """
+        In scenarios where you need to customize behaviour when invoking the \
+        generator, use this option to specify individual overrides. For example:
+
+        --option "entities.filenameTemplate: '%0DTO.swift'"
```
LePips, line 31: In other CLIs that provide similar functionality, they are typically comma separated key=value pairs and may fit in more in the context of a CLI. I'm more than happy to break from this convention as that has just been my experience with CLIs.

Additionally, are the double quotes for strings necessary? I think they can feel out of place in a CLI (again, my experience)

liamnichols: So the `--option` flag is repeatable here meaning that you don't need to comma separate anything.. To take the example from the tests, it ends up being something like this:

```bash
$ create-api generate \
    --option "entities.include: [Pet]" \
    --option "entities.mutableProperties: false" \
    --option "entities.filenameTemplate: 'Models.swift'" \
    --option "access: internal"
```

Wrapping in quotes is necessary (in some cases) because that whole string within is an argument. To do it without quotes, you'd have to make sure that there are no special characters such as whitespace or other quotes. But in some cases it's possible:

```bash
$ create-api generate --option access:internal
```

When you say comma separated `key=value`, do you mean something like this:

```bash
$ create-api generate --options "entities.include:[Pet], entities.mutableProperties=false, entities.filenameTemplate:Models.swift, access=internal"
```

So all of the overrides a technically a single string argument? 

---

Originally I thought that I had to put the entire `option:value` pair in a single command line argument (i.e `entities.filenameTemplate: '%0DTO.swift'`) because I didn't think Swift Argument Parser allowed dynamic arguments, however it looks like that might not be the case. I'll have a think to see if I can figure a way to separate them into separate arguments so that maybe something like this is possible:

```bash
$ create-api generate \
    --config create-api.yaml \
    --package MyPackage \
    schema.json \
    entities.include "[Pet]" \
    entities.mutableProperties false \
    entities.filenameTemplate Models.swift \
    access internal
```

Although I'm not sure how well the dot notation syntax as an option name feels ... I'll keep playing

LePips: My experience with CLIs are the following:
```
my-command --options firstKey=foo,secondKey=bar,thirdKey="Has Spaces"
```

As a reference, the OpenAPITools generator has this format: https://github.com/OpenAPITools/openapi-generator#3---usage.

My comment is strictly from my experience and I'm more than happy for the multi `--option` format.

liamnichols: Right I see what you are saying...

I think I'd prefer not to use comma separated list the option overrides can potentially be complex YAML values (I.e hashes or arrays) so its better to encourage them being separated into individual arguments. 

As for the `:` vs `=`.. I went with `:` since its familiar with both JSON and YAML but you are right, it does feel a bit out of place in a CLI... I could change it to be `=`, or maybe even support both. 

Finally, for the `--option` flag, there are two approaches with Swift Argument Parser.. Either I can have it as an array of `@Option`'s where you have to specify `--option` followed by the `key=value` pair, or I can make it an `@Argument` that parses all of the remaining arguments after the spec path... Each has pros and cons:

- Option: Pros are that you can place them in any order, cons are that its repetitive with a lot of overrides
- Argument: Pros are that you don't need to repeat `--option`, cons are that you have to make sure that they go after the input spec path 

Comparison of the usage:

```bash
$ create-api generate \
    --option "entities.include=[Pet]" \
    --config ".create-api.yml" \
    --option "entities.mutableProperties=false" \
    "schema.json" \
    --option "entities.filenameTemplate='Models.swift'" \
    --option "access=internal"
$ 
```

vs

```bash
$ create-api generate \
    --config ".create-api.yml" \
    "schema.json" \
    "entities.include=[Pet]" \
    "entities.mutableProperties=false" \
    "entities.filenameTemplate='Models.swift'" \
    "access=internal"
```

liamnichols: Oh also I just found the `.upToNextOption` parsing strategy which is probably the best of both worlds, it allows for this:

```bash
$ create-api generate \
    --options "entities.include=[Pet]" "entities.mutableProperties=false" "entities.filenameTemplate='Models.swift'" "access=internal" \
    --config ".create-api.yml" \
    "schema.json" \
```

Or you can continue to use `--options` multiple times like the first example as well

liamnichols: There is also `.remaining` that would allow for this:

```bash
$ create-api generate \
    --config ".create-api.yml" \
    "schema.json" \
    --options \
    entities.include "[Pet]" \
    entities.mutableProperties false \
    entities.filenameTemplate Models.swift \
    access internal 
```

Where basically anything after `--options` is parsed into an array that we can interpret however we want.. I could then join the key (i.e `entities.include`) arguments together with the values (i.e `"[Pet]"`) manually.

Maybe I am more a fan of this actually 🤔 

We could also support parsing single `key=value` argument pairs at the same time just like ArgumentParser currently allows us to do stuff like `--output=Generated` which is the equivalent to `--output` and `Generated` as separate arguments.

LePips: Currently, our CLI signature is the following:
```
Usage: create-api generate [<options>] <input>
```

I'm probably more of a fan also if `.remaining` since it's the largest list. This would change our signature to:
```
Usage: create-api generate [<options>] <input> [--options options]
```

However it's confusing with both `[<options>]` and ` [--options options]`, so we could rename `--options` to `--additional-options` or `--generate-options` to specify that these values control the generation.

Overall, I'm fine with the way that the key:value/key=value pairs are in strings and am for either key:value or key=value.

Additionally, since this allows a _very large_ list to be created it can be confusing with the schema input. So I also propose we change the signature to the following where the schema comes first and then all options are specified.

```
Usage: create-api generate <input> [<options>] [--options options]
```

liamnichols: > This would change our signature to:
> 
> ```
> Usage: create-api generate [<options>] <input> [--options options]
> ```

Not quite, `.remaining` is a parsing option for arguments so it would be this:

```
USAGE: create-api generate [<options>] <input> [<options> ...]
```

The difference is that you don't need to specify a `--` prefixed option name for arguments. If we used arguments, we'd have to mark sure that it goes after the spec input path.

> so we could rename `--options` to `--additional-options` or `--generate-options` to specify that these values control the generation.

How about `--config-options` to mentally associate it with `--config`? 

---

To summarise the possibilities

### Arguments (Remaining)

```swift
@Argument(parsing: .remaining)
var configOptions: [String]
```

```
USAGE: create-api generate [<options>] <input> [<config-options> ...]
```

```bash
$ create-api generate --package Foo schema.json access=internal entities.mutableProperties=false
```

### Options (default)

```swift
@Option
var configOption: [String]
```

```
USAGE: create-api generate [<options>] <input>
```

```bash
$ create-api generate --package Foo --config-option access=internal --config-option entities.mutableProperties=false schema.json
```

### Options (until next option)

```swift
@Option(parsing: .upToNextOption)
var configOptions: [OptionOverride]
```

```
USAGE: create-api generate [<options>] <input>
```

```bash
$ create-api generate --package Foo schema.json --config-options access=internal entities.mutableProperties=false
```

> **Note**: You don't have to put `--config-options` at the end, but if you do put it before `input`, you have to make sure that there is another option such as `--output` before it. So this is also possible:
>
> ```bash
> $ create-api generate --config .create-api.yml --config-options access=internal entities.mutableProperties=false --package Foo schema.json
> ```
>
> But not this:
>
> ```bash
> $ create-api generate --package Foo --config-options access=internal entities.mutableProperties=false schema.json
> ```
>
> That one would cause schema.json to be parsed as a config option override

---

After going through the options, while I think I do kind of like `@Option(parsing: .upToNextOption)`, I don't like that the generated USAGE doesn't make it clear that `<input>` will not be parsed properly in some cases. 

Maybe I'm back to square one of using just `@Option` with the default behaviour and repeatedly having to specify `--config-option` always using an `=` or `:` syntax (I'm leaning towards `=` as it feels more natural in cli arguments)

LePips: Thanks for laying out all the options! (pun not intended). Looking at these I think that `@Option` might just be the way to go for simplicity.

liamnichols: Ha, yeah.. Thanks for sense checking with me too. I'm going to make two adjustments:

1. Rename the option to `--config-option` to better associate it with `--config`
2. Update the format to use `key=value` instead of `key:value` to feel more aligned with cli practices

PR #128, opened 2022-08-08
Refactor CreateAPITests structure and snapshotting

liamnichols:
- cleanup for #123 

While working on the above PR, I had to shift some stuff around so I decided that it would be better to do some refactoring beforehand to get everything that I wanted into place. 

In this change, I do the following:

1. Introduce a new `Snapshotter` class that replaces the old `compare(expected:actual:)` global function.
2. Move test helpers (`TemporaryDirectory`) into **Helpers** directory
3. Remove generated snapshots, test specs and AllPackages out of **./Tests/CreateAPITests** and into a new **./Tests/Support** directory
    - I wasn't a big fan on having to manage excludes to keep the files from being compiled etc.
4. Move the `GenerateTestCase.SpecFixture` type into its own file at **./Helpers/SpecFixture.swift** now that we have some room to breathe with helpers 

Before|After
---|---
Before|After


LePips:
Occam's Razor, was gonna look at implementing the same thing.

PR #129, opened 2022-08-08
Introduce 'Record Snapshots' scheme

liamnichols:
Until now, I didn't realise that you could still share Xcode schemes when working with Swift Package projects. 

Knowing this means that we can avoid having to modify **Snapshotter.swift** every time that we want to re-record snapshots by instead having a shared scheme that defines the environment variable. 

To achieve this, I updated the **.gitignore** to match one that was produced by running `swift package init` (I added the custom ignore at the end of the file.

Screenshot 2022-08-08 at 18 21 11

cc @LePips

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,13 @@
+# Default ignores (run `swift package init` to compare)
 .DS_Store
-/.build
+.build
 /Packages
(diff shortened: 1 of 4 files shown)


LePips:
Ooh, fancy 🎉

Issue #130, opened 2022-08-08
`enum` forced unknown and organization


LePips:
## Forced unknown
Since we are creating API interfaces, new cases may be created by available from the server that aren't available on a current version of a developer's SDK or even their available schema. Additionally, only [Apple can create nonfrozen enums](https://docs.swift.org/swift-book/ReferenceManual/Statements.html) by my understanding so developers cannot use `@unknown default`. So, if an app with an out-of-date SDK attempts to decode an unknown case for a property or return type that cannot be optional, it will crash.

To fix this issue, create a new config option `enumForcedUnknown` that adds a new case to every enum which represents an unknown case. The name is open for improvement, but I think that it should stand out from typical Swift formatting in a way that ensures it will be different from any case in the schema **and** tells the developer that this is strongly unknown.
```swift
enum MyTypes {
    case foo
    case bar
    case forced_unknown // added unknown case
}
```

## Organization

Smaller pitch for an organization option to put `enum`s in a separate folder: `Enums`. Just as an organizational measure this can help a lot to separate the `struct/class` entities from other Swift types. This would only apply to the split generation.

LePips:
> The JSON Schema/OpenAPI specification does not support any differentiation between frozen or unfrozen enums and technically when you are using a different schema, it should be versioned differently.

No, that's because this lands in the territory of how a Swift developer uses this schema. This isn't about frozen/unfrozen enums _within_ the same version of the schema but across different schema versions. I still very much consider this "unfrozen" as an API is very much open to evolution. It's somewhat obvious that an application should not crash due to unexpected responses from an API, at least as much as it can.

---
Basic example, if `Pet.type` was of type (not optional) enum with cases `.dog, .cat` in V1 and then V1.1 added `.bird` to the available cases:

1 - if the developer didn't update the app's API to V1.1 it would crash for everyone
2 - if the developer did update the app's API but people haven't updated the app, it will crash for them

Instead the application just probably wouldn't show these unknown types and they can only be shown to people who update their app.

---

We don't necessarily have the responsibility to handle this, after all that is what the schema _explicitly_ states, but it would definitely help us developers _use_ an API. This kind of addition (among all the other options like exclusions and renamings I argue) is strictly Swift developer facing and does not actually go _against_ the schema, it just helps us work with safety and convenience.

> `RawRepresentable` instead of enum

[...]

Issue #85, opened 2022-08-04
Support inclusion of custom package dependencies


liamnichols:
# Background

While it's possible to define additional `import`'s in the generated paths or entities code, it is not yet possible to declare dependencies to those modules in the generated **Package.swift**.

When investigating how to give the user more control over importing Get (and I guess other default dependencies), it became a bit clearer that we needed a more powerful way to define which dependencies we want to use and what they bring to the generated code. 

Se notes here: https://github.com/CreateAPI/CreateAPI/issues/84#issuecomment-1205026288

Fundamentally, I believe that we need a way to declare custom dependencies, and with that:

- Where the dependency comes from and it's version requirements (if it needs to go into **Package.swift**)
- What the dependency provides (i.e the `Request` type)
- Where the custom dependency is required (i.e maybe it only needs improving into entities)

With this information, we'd able to do the following:

1. Merge the custom dependencies with required dependencies (i.e Get) where appropriate
2. Figure out which files need to import which module
3. Embed each dependency in **Package.swift**, including the custom ones

# Examples

### Previous Behaviour

For matching the previous behaviour, the following config:

```yaml
entities:
  imports:
  - CoreLocation
```

Would become this:

```yaml
dependencies: 
- module: CoreLocation
  importIn:
  - entities
```

We can deprecate the old options and automatically backfill them during a transition period.

### Package Dependencies

In the event that we also need to bring in a package to help with some additional customisations:

```yaml
entities:
  protocols: 
  - DataTransferObject
dependencies:
- module: DataTransferKit
  url: https://github.com/myorg/DataTransferKit.git
  from: "0.0.2" # need other options too, such as branch, revision, upToNextMinor etc
  importIn:
  - entities
```

This will both import `DataTransferKit` in generated entity files for the protocol conformance, and will also include the dependency in the generated Package.swift file.

### Overrides

Finally, being able to override the existing imports. For example, Get:

```yaml
dependencies:
- module: MyAPIKit
  url: https://github.com/myorg/MyAPIKit.git
  branch: main
  provides:
  - Request
```

Instead of using `importIn`, this tells CreateAPI that this dependency provides the given type which would allow it to replace built-in dependencies that provide that type (Get). 

Right now, the generator would just import `MyAPIKit` instead of `Get` in all Paths, but in the future we would have the potential to expand this to any type. For example:

```yaml
paths:
  overriddenResponses:
    MyEntity: GenericType
dependencies:
- module: MyUtilsKit
[...]

PR #134, opened 2022-08-10
Move generator related CLI options into configuration file

### LINKED ISSUE #47: All command line arguments that alter output should be configurable in config file

liamnichols:
- Closes #47 
- https://github.com/CreateAPI/CreateAPI/issues/130#issuecomment-1214136925

The `--merge-sources`, `--package`, `--module`, `--vendor`, `--generate` and `--entityname-template` options all alter the generator output but they are currently passed separately to the rest of the configuration options.

As part of the linked ticket, we want all options that alter the behaviour of the generator to be sourced from the `ConfigOptions` type which loads primarily from the **.create-api.yml** configuration file. Doing so would allow the user to store all important information in a single place without having to worry about making sure that these values are always passed when they invoke the cli by wrapping in a script etc.

In cases where the user might actually want to pass these kind of options, or others that were previously only defined in the configuration file, we've introduced the general purpose `--config-option` option that allows you to pass a `key.path=value` pair. See #126 for more info.

Refer to the table below for a mapping of the options

Before|After
---|---
`--merge-sources`|`mergeSources`
`--package`|`generate` and `module`
`--module`|`generate` and `module`
`--vendor`|`vendor`
`--generate`|`generate`
`--entityname-template`|`entities.nameTemplate`

For more information, keep reading

### `mergeSources`

This one is pretty straightforward, it's just a boolean flag at the top level:

```yaml
mergeSources: true 
```

The default value is `false` 

### `generate` and `module`

These two options differ slightly to how they were previously used since they are no longer mutually exclusive. `generate` is now a set with three options (`paths`, `entities`, and `package`) and `module` is the name of the module being generated. 

`module` is required and `generate` defaults to `[entities, paths, package]`. So by default a complete package is generated, and all you need is to specify the module name:

```yaml
module: MyCustomAPIKitInAPackage
```

If you don't want the package, you'd do something like this:

```yaml
module: MyAPIKitInAPackage
generate: [entities, paths, enums]
```

Note that `isGeneratingEnums` has also been incorporated into the `generate` option.

### `vendor`

Pretty straightforward although this is now an `enum` so you can't specify anything other than `github` or null.

```yaml
vendor: github
```

### `entities.nameTemplate`

Finally, the `--entityname-template` option has moved into `entities` like so:

```yaml
entities:
  nameTemplate: %0DTO
```

---

Feedback on the approach is welcome, because there may well be a better way to approach this.


Tests/CreateAPITests/Compliling/PackageManifest.swift
```diff
@@ -45,6 +45,26 @@ struct PackageManifest {
...
+            return nil
+        }
+
+        let groupRange = match.range(at: 1)
+        guard let range = Range(groupRange, in: contents) else {
+            return nil
+        }
+
+        return String(contents[range])
+    }
```
liamnichols, line 67: Previously when running snapshot tests we could just read the `package` option from the parsed `Generate` command but that is no longer the case because the config file (or overrides) aren't loaded until after running the generator and there is no way to pass that information back out.

Instead, I opted to read the package name from the generated **Package.swift** file if it exists. 

There might be a better approach for this in the future but it should do the trick for now.

Tests/CreateAPITests/GenerateFeaturesTests.swift
```diff
@@ -5,10 +5,7 @@ final class GenerateFeaturesTests: GenerateTestCase {
     func testQueryParameters() throws {
         try snapshot(
             spec: .testQueryParameters,
-            name: "test-query-parameters",
-            arguments: [
-                "--package", "test-query-parameters"
-            ]
+            name: "test-query-parameters"
```
liamnichols, line 8: In `GenerateTestCase`'s `snapshot` method, I've updated it to infer the `moduleName` from the `name` parameter. This means that in a lot of places we no longer need to pass the `moduleName` so you'll see that `--package` turns into nothing in the diff.

Tests/CreateAPITests/Compliling/PackageManifest.swift
LePips: The folder name: `Compliling` -> `Compiling`

LePips:
Continued from https://github.com/CreateAPI/CreateAPI/issues/85.

---

- The `package.dependencies` structure sounds good, so my PR can wait until this one is merged. I think it wouldn't hurt to have `packages.name` as it's a mixed bag for whether the package name should match a module name in a lot of SPM packages.
- The `generate: entities, paths` also looks good
- I don't think that the concept of `enabled` fits in a config: options specified in a config should happen (unless overridden or incorrect) and I don't fancy the idea of "disabling" a section if it was written out. So, I prefer the above option of `generate`.

liamnichols:
> I think it wouldn't hurt to have `packages.name` as it's a mixed bag for whether the package name should match a module name in a lot of SPM packages.

Yeah, in most cases you typically name the package after the name of the product that it exports (which is the target name) and we do this already. But there are cases where it differs. 

For example, the swift-argument-parser package has a target called ArgumentParser. If `package.name` is unset, we can continue with the legacy behaviour 👍 

I won't tackle that in this PR though

liamnichols:
@LePips: I pushed up the changes, since I'm removing the `enabled` flags for the single `generate` and am not adding `package.name` here, I've not actually added the `package` option so you can add it directly in your PR. 

I can add `package.name` afterwards 👍

liamnichols:
As per https://github.com/CreateAPI/CreateAPI/issues/130#issuecomment-1214136925, I've also included `enums` as part of the `generate` option set.

Issue #114, opened 2022-08-05
Support custom mappings between schema formats and Swift types


LePips:
Edit: differs from the original to a single global override instead of individual properties and I forgot parameter types. I think that we can start with a global override and somebody can request this fine tuning later.

---

Introduce a config option to override the Swift type when they are parsed and written:

```yaml
# schema
Pet:
  properties:
    id:
      type: string
    tag:
      type: string
      format: uuid

# config
swiftTypeOverrides:
  UUID: String # any occurrences of UUID are replaced with String, so `Pet.tag: String`
```

Config name open for improvement. Obviously, a developer might override to a type that cannot be decoded from/encoded to the original type, like `UUID: Int` or a type that does not exist, like `UUID: qwerty`. So, it is a developer responsibility to make sure that overriding types can be properly decoded/encoded and that they are a proper Swift type.

---

This can be further enhanced with imports as we should be trivially able to set an imported type after https://github.com/CreateAPI/CreateAPI/issues/85 is figured out. Using NaiveDate as an example removing the `useNaiveDate` option, we include the package as an import in the config and then set the type override:

```yaml
# schema
Pet:
  properties:
    birth_date:
      type: string
      format: date

# config
swiftTypeOverrides:
  Date: NaiveDate # any occurrences of Date are replaced with NaiveDate
```

Obviously, imported objects should be expected to properly decode/encode but that's an implementation detail of the imported type.

liamnichols:
To just clarify the scope here, I don't think we want to override the Swift types used, but more that we want the ability to customise how the `format` of a type should be interpreted? 

If we summarise what we have today, its somewhat like the following:


```yaml
string:
  date: Date
  date-time: NaiveDate # if useNaiveDate
  uuid: UUID
  uri: URL
  byte: Data
integer:
  int32: Int32 # if useFixWidthIntegers
  int64: Int64 # if useFixWidthIntegers
number:
  float: Float
  double: Double
```

With the current `type`/`format` -> Swift type behaviour, there are a few drawbacks:

1. It's spread across multiple configurations
2. We don't provide handling for even some of the defined formats (i.e `password`) 
3. There is no way to customise/override beyond `useNaiveDate` and `useFixWidthIntegers`

So what we want is for the generator to be able to use alternate mappings in addition to the default ones? 

[...]

PR #135, opened 2022-08-11
Remote Dependency Support

LePips:
- Part of https://github.com/CreateAPI/CreateAPI/issues/85

Adds the ability for SPM dependencies to be injected into the Package manifest. The config syntax looks like below:
```yaml
dependencies:
- url: https://github.com/apple/swift-distributed-actors
  module: DistributedActors
  rule:
    commit:
      hash: 9da657e933f7c302277eef373c4a0d74e90f2bfa
```

All rules allowed in SPM are included. This greatly enhances the `_.imports` functionality and contributes to the enhanced functionality of https://github.com/CreateAPI/CreateAPI/issues/114.

Documentation:
- https://github.com/LePips/CreateAPI/blob/remote-spm-dependencies/Docs/ConfigOptions.md#dependencies

It's quite simple but shows all of the rules. I thought about making an enum for all of the `rule` type options however I think that in a list format it can be hard to layout so I have examples of all of them. The general description is quite straightforward.

---

The declaration of a rule can theoretically be simplified to below, especially since most options (except the ranges) require a more simpler declaration.
```yaml
rule:
  commit: 9da657e933f7c302277eef373c4a0d74e90f2bfa

# but ranges would be unchanged
rule:
  range:
    from: 1.0.0
    to: 3.1.2
```

I started out aiming for this syntax however from a development (and therefore maintenance) perspective it was too complicated and heavy due to the decoding and using objects instead of an enum. What is implemented now is the simplest and another implementation would save only a single line per import, a negligible optimization.

There is practically no error checking regarding URLs or version types as a representable wrapper doesn't decode properly. So, this puts responsibly on the developer to make sure that inputs are correctly formatted. 

We could possibly look at a lot more robust config validator. I've been thinking about that for printouts, so it may need to come from the config parser project.


Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,58 @@
...
+    let url: URL
+    let module: String
+    var rule: DependencyRule
+    
+    public var packageDeclaration: String {
+        ".package(url: \"\(url)\", \(rule.declaration))"
+    }
+    
+    public var productDeclaration: String {
+        let cleanedPackage = url
```
LePips, line 46: This is done per [SPM 0292](https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md).

LePips:
Many of the existing packages were updated due to some trailing whitespacing being eliminated.

Docs/ConfigOptions.md
```diff
@@ -308,6 +309,65 @@ commentOptions:
 
 <br/>
 
+## packages
+
+**Type:** [PackageDeclaration]<br />
+**Default:** `[]`
+
+Additional remote Swift Package imports.
```
LePips, line 317: I was going to add more documentation to say that you can use this alongside `entities/paths.imports` but that can wait till https://github.com/CreateAPI/CreateAPI/issues/114 so that it's even more descriptive.

Sources/CreateAPI/Generator/Generator+Package.swift
```diff
@@ -8,19 +9,19 @@ extension Generator {
...
-            #".package(url: "https://github.com/kean/Get", from: "1.0.2")"#,
-            isHTTPHeadersDependencyNeeded ? #".package(url: "https://github.com/CreateAPI/HTTPHeaders", from: "0.1.0")"# : nil,
-            isNaiveDateNeeded ? #".package(url: "https://github.com/CreateAPI/NaiveDate", from: "1.0.0")"# : nil,
-            isQueryEncoderNeeded ? #".package(url: "https://github.com/CreateAPI/URLQueryEncoder", from: "0.2.0")"# : nil
-        ].compactMap { $0 }.joined(separator: ", \n")
+        let allPackages = options.package.dependencies
+            .appending(GeneratorPackages.Get)
+            .appending(if: isHTTPHeadersDependencyNeeded, GeneratorPackages.HTTPHeaders)
+            .appending(if: isNaiveDateNeeded, GeneratorPackages.NaiveDate)
+            .appending(if: isQueryEncoderNeeded, GeneratorPackages.URLQueryEncoder)
```
liamnichols, line 16: This is so much better 😍 

A couple of suggestions:

1. Define define the constants on `PackageDeclaration` so that we can drop the repetition
2. Constants should still use the `lowerCamelCase` syntax 
3. For the function to read a little more naturally, maybe `if` could be the second argument? 

For example:

```suggestion
        let allPackages = options.package.dependencies
            .appending(.get)
            .appending(.httpHeaders, if: isHTTPHeadersDependencyNeeded)
            .appending(.naiveDate, if: isNaiveDateNeeded)
            .appending(.urlQueryEncoder, if: isQueryEncoderNeeded)
```

Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+public enum DependencyRule: Decodable {
+    case exact(version: String)
+    case upToNextMajor(from: String)
+    case upToNextMinor(from: String)
```
liamnichols, line 6: I propose that we drop these since they are deprecated in Swift 5.6 with the intention that you specify the range yourself. Once our Package.swift targets 5.6 above, we'd have to manually deconstruct the string and figure it out ourselves ([like this](https://github.com/apple/swift-package-manager/blob/8970e016e9688f0e1fca549cdc02719fa5359059/Sources/PackageDescription/PackageRequirement.swift#L216-L231)) which isn't worth it. 

If we don't include it now, it'll encourage people to use `range` or `from` directly which will translate into a Swift 5.6 friendly format once that item comes (probably after 5.7 is publicly released so in two months)

Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,58 @@
...
+        case .closedRange(from: let min, to: let max):
+            return "\"\(min)\"...\"\(max)\""
+        case .branch(name: let name):
+            return ".branch(\"\(name)\")"
+        case .commit(hash: let hash):
+            return ".revision(\"\(hash)\")"
+        case .from(let version):
+            return "from: \"\(version)\""
+        }
+    }
```
liamnichols, line 32: For reference form my last comment, when we change the package-tools-version to 5.6, this will become the following:

 ```suggestion
    var declaration: String {
        switch self {
        case .exact(version: let version):
            return "exact: \"\(version)\""
        case .range(from: let min, to: let max):
            return "\"\(min)\"..<\"\(max)\""
        case .closedRange(from: let min, to: let max):
            return "\"\(min)\"...\"\(max)\""
        case .branch(name: let name):
            return "branch: "\(name)\""
        case .commit(hash: let hash):
            return "revision: "\(hash)\""
        case .from(let version):
            return "from: \"\(version)\""
        }
    }
```

File: Sources/CreateOptions/PackageDeclaration.swift [hunk 6]
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+public enum DependencyRule: Decodable {
```
liamnichols, line 3: This is referred to as a `Requirement` by SPM, maybe let's match the terminology for consistency?

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -537,4 +537,61 @@ final class GenerateOptionsTests: GenerateTestCase {
...
+                - url: https://github.com/apple/swift-atomics
+                  module: Atomics
+                  rule:
+                      upToNextMajor:
+                          from: 1.0.0
+                - url: https://github.com/apple/swift-syntax.git
+                  module: SwiftSyntax
+                  rule:
+                      upToNextMinor:
+                          from: 0.50600.1
```
liamnichols, line 566: I'm not sure about the about the other examples but I think they are fine, but can we switch SwiftSyntax out for something a little more lightweight since CI has to clone and compile this package and its dependencies. 

That being said, if it doesn't take too much time then it's fine but last I remember, it was big 😅

Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,58 @@
...
+        case .from(let version):
+            return "from: \"\(version)\""
+        }
+    }
+}
+
+public struct PackageDeclaration: Decodable {
+    
+    let url: URL
+    let module: String
```
liamnichols, line 38: I'm not sure what is the most appropriate term to use here. After reading through https://forums.swift.org/t/module-vs-product-vs-target/27640, I have a feeling that what we actually want is 'product'.

Additionally, maybe this needs to be an array in scenarios where you need multiple products offered by the package? 

For example:

```yaml
package:
  dependencies:
  - url: https//github.com/apple/swift-syntax
    from: 0.50600.1
    products:
    - SwiftSyntax
    - SwiftSyntaxParser

File: Sources/CreateOptions/PackageDeclaration.swift (hunk 1 above)
Comment (liamnichols) [reply Sources/CreateOptions/PackageDeclaration.swift:46]: I don't think its down to us to clean this though? That should be on the user right? Just like we don't validate/clean the versioning strings?

**Review by liamnichols — COMMENTED** [review]
This is looking great! Good job on such a useful change 🙌

File: Sources/CreateOptions/PackageDeclaration.swift (hunk 1 above)
Comment (liamnichols) [reply Sources/CreateOptions/PackageDeclaration.swift:46]: Sorry, I missed how/why we are doing the cleaning.. This is for the package name and not the URL itself, makes perfect sense!

File: Sources/CreateOptions/PackageDeclaration.swift [hunk 9]
```diff
@@ -0,0 +1,58 @@
...
+    var rule: DependencyRule
+    
+    public var packageDeclaration: String {
+        ".package(url: \"\(url)\", \(rule.declaration))"
+    }
+    
+    public var productDeclaration: String {
+        let cleanedPackage = url
+            .lastPathComponent
+            .split(separator: ".")[0]
```
liamnichols, line 48:
```suggestion
        let cleanedPackage = url
            .deletingPathExtension()
            .lastPathComponent
```

Sources/CreateOptions/PackageDeclaration.swift
LePips: Xcode uses the terminology "Dependency Rule" so I just matched it.

liamnichols: Interesting 😄 it's called a `SourceControlRequirement` in the SPM source:

https://github.com/apple/swift-package-manager/blob/8970e016e9688f0e1fca549cdc02719fa5359059/Sources/PackageDescription/PackageRequirement.swift#L135-L191

For comparison, [XcodeGen uses 'Version'](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#remote-package) and [Tuist appears to also use Requirement](https://github.com/tuist/tuist/blob/eede714adf9cf3910b6750f1efd439aff35515a8/Sources/ProjectDescription/Package.swift#L18-L35)

I think in this case, it's still better to match the SPM terminology to favour the actual source (and be a bit more considerate of Swift outside of Xcode) 🙇

Sources/CreateOptions/PackageDeclaration.swift
LePips: I actually did use an array but with testing, but by my testing when a package has dependencies on another package it imports all of the products? I don't know the specifics.

By my testing, I had SwiftNumerics as a dependency as only had Numerics as a product. I imported the generated package and was able to use RealNumerics and ComplexModule.

Tests/CreateAPITests/GenerateOptionsTests.swift
```diff
@@ -459,6 +459,51 @@ final class GenerateOptionsTests: GenerateTestCase {
             """
         )
     }
+    
+    func testPetstoreSPMImports() throws {
+        try snapshot(
+            spec: .petstore,
+            name: "petstore-SPM-imports",
+            testCompilationOnLinux: false, // custom imports aren't available there.
```
liamnichols, line 467: Any chance that we can test with linux friendly dependencies?

Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -232,6 +232,58 @@ public struct ConfigOptions: ParsableConfiguration {
...
+        ///     rule:
+        ///       range:
+        ///         from: 1.0.0
+        ///         to: 2.1.0
+        ///   - url: https://github.com/apple/swift-metrics.git
+        ///     product: Metrics
+        ///     rule:
+        ///       closedRange:
+        ///         from: 2.0.0
+        ///         to: 3.1.0
```
liamnichols, line 267: I can never remember the difference between range or closed range... Could these examples maybe better reflect how you might use them?

```suggestion
        ///   - url: https://github.com/apple/swift-algorithms
        ///     product: Algorithms
        ///     rule:
        ///       range:
        ///         from: 1.0.0
        ///         to: 2.0.0
        ///   - url: https://github.com/apple/swift-metrics.git
        ///     product: Metrics
        ///     rule:
        ///       closedRange:
        ///         from: 2.0.0
        ///         to: 2.9.1
```

I think that is the right way? 😅 So `closedRange` would also pick 2.9.1 if it was available whereas `range` wouldn't pick 2.0.0?

Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,68 @@
...
+            return "from: \"\(version)\""
+        }
+    }
+}
+
+public struct PackageDeclaration: Decodable {
+    
+    let url: URL
+    let product: String
+    var rule: SourceControlRequirement
```
liamnichols, line 33: We could also change the property name to `requirement`?

Sources/CreateOptions/PackageDeclaration.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+
+public enum SourceControlRequirement: Decodable {
+    case exact(version: String)
+    case range(from: String, to: String)
+    case closedRange(from: String, to: String)
+    case branch(name: String)
+    case commit(hash: String)
+    case from(version: String)
```
liamnichols, line 9: I guess we are getting `Decodable` for free here by labelling the associated values? 

It would be nice if we could avoid having to nest the value for cases where there is only one associated value. For example, if this:

```yaml
- url: https://github.com/apple/swift-argument-parser
product: ArgumentParser
rule:
  exact:
    version: 1.1.1
- url: https://github.com/apple/swift-log
product: Logging
rule:
  branch:
    name: main
- url: https://github.com/apple/swift-numerics
product: Numerics
rule:
  commit:
    hash: 7f2d022d3d9b55bf812814f5d01896cbfa0fd4da
- url: https://github.com/apple/swift-system
product: SystemPackage
rule:
  from:
    version: 1.2.1
```

Could become this:

```yaml
- url: https://github.com/apple/swift-argument-parser
product: ArgumentParser
rule:
  exact: 1.1.1
- url: https://github.com/apple/swift-log
product: Logging
rule:
  branch: main
- url: https://github.com/apple/swift-numerics
product: Numerics
rule:
  commit: 7f2d022d3d9b55bf812814f5d01896cbfa0fd4da
- url: https://github.com/apple/swift-system
product: SystemPackage
rule:
  from: 1.2.1
```

No problem if not though

liamnichols (commented):
🚀

Sources/CreateOptions/ConfigOptions.swift
LePips: Range would be exclusive of the upper bound and closed range inclusive would be inclusive of the upper bound, so yes it works as you think it would.

Sources/CreateOptions/PackageDeclaration.swift
LePips: We do get decodable for free however when an enum is decoded with an associated value (without name) it has to be of the format:
```yaml
rule:
  from:
	_0: 1.1.1
```

Please see the bottom of the PR summary for why this had to happen.

Sources/CreateOptions/PackageDeclaration.swift
LePips: My testing from within the package is consistent with what I tested before with an external project, we should only have to label one product and all other products are imported as well.

Sources/CreateOptions/PackageDeclaration.swift
liamnichols: > I started out aiming for this syntax however from a development (and therefore maintenance) perspective it was too complicated and heavy due to the decoding and using objects instead of an enum. What is implemented now is the simplest and another implementation would save only a single line per import, a negligible optimization.

Sorry yeah I think I read that but then forgot about it 😄 Lets keep the default Decodable implementation! Thanks

Sources/CreateOptions/ConfigOptions.swift
liamnichols: Great thanks. Could we commit the above suggestion to the docs to help distinguish the two a little more in that case please?

Sources/CreateOptions/PackageDeclaration.swift
liamnichols: It was probably cached in derived data/the build directory. Try cleaning the project and it won't work (or `swift build -c release`), you have to specify all products that are required so that it knows what to link.

LePips: Apologies, I got confused because a target will also import its dependencies and `Numerics` imported `RealNumerics` and `ComplexModule` as well.

LePips: Do we think we can generalize `OptionSetDecodable` so that the wrapped type doesn't have to be `CaseIterable`? It would help with the array syntax at least a bit as I assume most imports will probably use single products.

liamnichols: I don't think we can so much generalize it (or at least I didn't figure out a way to), but we could just add a similar `SingleElementDecodable` property wrapper instead?

Sources/CreateAPI/Generator/Generator+Package.swift
```diff
@@ -8,19 +9,22 @@ extension Generator {
...
-            isHTTPHeadersDependencyNeeded ? #".product(name: "HTTPHeaders", package: "HTTPHeaders")"# : nil,
-            isNaiveDateNeeded ? #".product(name: "NaiveDate", package: "NaiveDate")"# : nil,
-            isQueryEncoderNeeded ? #".product(name: "URLQueryEncoder", package: "URLQueryEncoder")"# : nil
-        ].compactMap { $0 }.joined(separator: ", \n")
+        let dependenciesDeclaration = allPackages
+            .map(\.productDeclarations)
+            .reduce(Array<String>()) { partialResult, currentImports in
+                return partialResult.appending(contentsOf: currentImports)
+            }
+            .joined(separator: ",\n")
```
liamnichols, line 27: I think you can also just `flatMap(_:)` the nested `[[String]]` array instead of having to use `reduce`

```suggestion
        let dependenciesDeclaration = allPackages
            .map(\.productDeclarations)
            .flatMap({ $0 })
            .joined(separator: ",\n")
```

liamnichols (approved):
🔥 Thanks for taking on all of the feedback 🚀 

Let me know if you wanted to followup on https://github.com/CreateAPI/CreateAPI/pull/135#discussion_r945329822 as part of this PR or if you wanted to leave it for some other time. 

I can merge whenever you are ready 👍

LePips:
Thanks for the feedback! I'm fine for the last comment to be done at a later time, merge away!

PR #136, opened 2022-08-11
Data Type Override

LePips:
- Depends on https://github.com/CreateAPI/CreateAPI/pull/135
- Closes https://github.com/CreateAPI/CreateAPI/issues/114

## What
As described in #114, a Swift type override can be very powerful when working together with `packages.dependencies` and `entites/paths.imports`. The encoding/decoding caveats and developer responsibilities are in the issue.

This PR differs in the format now being:
```yaml
dataTypeOverrides:
  data-type.format: SwiftType
```

The power comes from `format` being an open value on numbers, integers, and strings. Users can now use different built in types or create and import their own custom types for corresponding labels or a global override.

Examples:
- `string.uuid: String`
- `number.float: Float80`
- `integer.Int64: Int`, assuming `useFixWidthIntegers = true` _and_ one wants to use `Int32/Int`
- `string.password: MyPasswordType`, assuming custom import
- `string: MyStringType`, assuming custom import

Using imports would require at least the following:
```yaml
dataTypeOverrides:
  string.password: MyPasswordType
entities:
  import: MyPasswordType
packages:
  dependencies:
  - url: # MyPasswordTypeURL
  # rest of the dependency configuration
```

Again, it is a developer responsibility that these types are properly encoded/decoded, that they exist, and that it is proper within the context of their API.

## TODO
- [x] Remove `useNaiveDate`
	- Leaving in for this implementation.

However, we can recommend NaiveDate and other packages that can enhance the experience in the documentation!
- [x] Interplay with `useFixWidthIntegers`.
    - ~~Should this or overrides take precedence?~~ Consensus reached to remove this and width types can be set here.
- [ ] Documentation
    - Overall example usage, noting that this is **Advanced** configuration.
- [x] Dictionary/array types (edited after some thought)
    - These are not "types" but "objects". This can be a future feature request. There can be a use case for this where:
    1 - `string.first_name: MyFirstNameType`, custom format `first_name`
    2 - An array object in the schema represents an array of `first_name`-formatted strings. So we want `[MyFirstNameType]`. This isn't configurable in the schema so this is something that developers should know within the context of their API.
- [ ] Split per `entities/paths`?
    - A use case I just thought of is where somebody would want to use their entities with something like `string.uuid: String` for convenience _but_ when making calls that would then require an entity UUID as a parameter they want to enforce at runtime that the type is correct. This would increase to possibly three flags: `globalDataTypeOverrides`, `entites.dataTypeOverrides`, and `paths.dataTypeOverrides`. I say we ignore this and can be an enhancement later on if requested as it's somewhat similar to the individual property type override idea from the original issue.


liamnichols:
I still need to spend some time going over these proposed changes as well as #114 in more detail, but from skimming over the description I just want to mention that removing NativeDate in favour of the configuration proposed is probably not the direction that we want to go in. 

We offer NaiveDate (and HTTPHeaders and URLQueryEncoder) as sensible defaults that help fill the gap where Swift/Foundation can't quite represent common parts or behaviours of the OpenAPI output. These features should remain opt-out rather than opt-in.

I think that we should have a better way to represent all of these default dependencies, but having to use the more advanced configuration to opt-in isn't preferable here

LePips:
I originally commented about the distinction between NaiveDate and HTTPHeaders/URLQueryEncoder where the latter are used by the generator. These are fine to be included for our use. NaiveDate is _purely_ for developer use to ignore the timezone for dates, so it's  just a helper package we provided behind an option. It doesn't fill in any gaps per say for Swift/Foundation.

> can't quite represent common parts or behaviours of the OpenAPI output

Then the schema would be incorrect. In the context of dates, it wouldn't be a proper RFC 3339 date and that's a developer concern.

> but having to use the more advanced configuration to opt-in isn't preferable here

These do require some boilerplate, but in tandem with these other options it is a lot more powerful than putting things behind flags.

Overall, NaiveDate was just a package we included as convenience and now fits of just being a normal external dependency.

liamnichols:
> Then the schema would be incorrect. In the context of dates, it wouldn't be a proper RFC 3339 date and that's a developer concern. Swift is natively capable of representing all the data types per OpenAPI 3.*

That's not the case. Foundation's builtin `Date` type represents date and time but when using `format: date`, as per RFC 3339 you might not be including time or timezone information that is necessary to properly interpret the value as a `Date`. That gives us two options:

1. Use `String`
2. Use `NaiveDate`

We choose the latter by default because `date` is a format that is recognised/defined by the OpenAPI spec and its a common scenario that we'd like to be able to support:

https://swagger.io/docs/specification/data-models/data-types/#string

> An optional format modifier serves as a hint at the contents and format of the string. OpenAPI defines the following built-in string formats:
> - `date` – full-date notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, `2017-07-21`
> - `date-time` – the date-time notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, `2017-07-21T17:32:28Z`
> - `password` – a hint to UIs to mask the input
> - `byte` – base64-encoded characters, for example, `U3dhZ2dlciByb2Nrcw==`
> - `binary` – binary data, used to describe files (see [Files](https://swagger.io/docs/specification/data-models/data-types/#file) below)

In addition to the built-in formats:

> However, format is an open value, so you can use any formats, even not those defined by the OpenAPI Specification, such as:
> - `email`
> - `uuid`
> - `uri`
> - `hostname`
> - `ipv4`
> - `ipv6`
> - and others


We process `uuid` and `uri` because Swift has builtin types `UUID` and `Data` that matches in most cases. 

I do see your point though about how some of our existing flags such as `useFixWidthIntegers` and `useNaiveDate` make it a bit tricky to then be able to provide complete flexibility. 

I'm going to have a bit more of a read/look through this PR and also the original ticket and I'll followup some more after 👍

LePips:
I'll leave `useNaiveDate` in for this implementation, but I think a better implementation in the future would be decoding from a `DateFormatter` with format `yyyy-dd-MM` for `date-time`.

liamnichols:
> > Use `string`, use `NaiveDate`
> 
> We actually just set to `Date` if `useNaiveDate` is false for `date-time`.

That's not the case 😄 

https://github.com/CreateAPI/CreateAPI/blob/674d9d8b69542d9c00c41d95ed589fc1da6d43dc/Sources/CreateAPI/Generator/Generator%2BSchemas.swift#L232-L248

This switch statement could be refactored a little for clarity, but the gist of things is this:

- On line 234, if the format is `date-time`, it returns `Date`
- On line 237, if the format is **`date`** and `options.useNaiveDate` is `true`, it returns `NaiveDate`
- Otherwise, on line 247 it will return `String`

> but I think a better implementation in the future would be decoding with a `DateFormatter` with format `yyyy-dd-MM` for `date-time` instead of relying on an external package so that `Date` is still used.

It's worth checking this blog out for a bit more context on why `Date` isn't suitable for representing `date` (not `date-time`) formats - https://kean.blog/post/naive-date-time

---

With the approach that I suggested in #114, `useNaiveDate` would become redundant because the implementation of the above method would end up being something like this:

```swift
private func getStringType(for info: JSONSchema.CoreContext<JSONTypeFormat.StringFormat>) -> TypeIdentifier {
    let identifier = options.stringType(for: info.format)
    if identifier == .builtin("NaiveDate") {
        setNaiveDateNeeded()
    }
    return identifier
}
```

Maybe theres a better way to achieve it, but the idea being that you change the data type mappings in the options if you want to disable NaiveDate from being used.

LePips:
Apologies, my last comment did get `date/date-time` mixed up and I must've been confused and thought it also returned `Date` when `useNaiveDate` was false. I've been making this a bigger issue than it really is and I apologize for that as well.

I still think `useNaiveDate` is just a basic type replacement and could be removed since one could import it manually and use type replacement. However, that's just a creative difference. In my last comment I mentioned the date decoding style for native `Date` decoding for this case and think that would be the best in the long run, instead of depending on an external package or getting a `String`.

liamnichols:
Don't worry about it, it's a confusing and complicated topic 🙂 There are still reasons why we can't just use a different date formatter to decode `date` formatted strings (not `date-time`) but lets leave that for a separate discussion 👍 

If we focus here on formats preserving the existing set of default mappings then we can get this shipped 🚀

LePips:
I will close this PR due to the long unnecessary discussion from me (apologies again) above with an updated and more accurate PR.

PR #137, opened 2022-08-13
Change default behaviour of `isInliningPropertiesFromReferencedSchemas` (now `inlineReferencedSchemas`)

### LINKED ISSUE #63: Consider making `isInliningPropertiesFromReferencedSchemas` the default behaviour
# Background 

Context: https://github.com/CreateAPI/CreateAPI/pull/61#discussion_r931894606

When `isInliningPropertiesFromReferencedSchemas` is not enabled (the default), it generates an invalid `Encodable` implementation. But generally speaking, the nested objects seem to not be consistent with how other generators generate code and it leads me to think that it's not the right choice to be nesting entities. 

My proposal is that we remove the `isInliningPropertiesFromReferencedSchemas` option and make it's behaviour the default.

liamnichols:
# Background 

- Closes #63 

Currently CreateAPI takes one of two approaches when dealing with schemas that reference other types using `allOf`:

1. Nest the referenced entity within
2. Inline the properties from the referenced entity

Until this changes, the former was the default behaviour however this behaviour isn't really inline with what you might expect a generator to do and it results in a strangely formed (but functional) entity. 

In the future, we plan to improve the inlining behaviour by adding other alternatives (i.e protocol composition and maybe even class inheritance) but for now we want to change the default behaviour. 

# Changes

In this PR, I update `isInliningPropertiesFromReferencedSchemas` (now `inlineReferencedSchemas`) so that properties are inlined into the entity by default. 

I also regenerate the snapshots to reflect the change in default behaviour.

### DIFF
--- a/Docs/ConfigOptions.md
+++ b/Docs/ConfigOptions.md
@@ -559,7 +559,58 @@ If set to `true`, uses the `default` value from the schema for the generated pro
 **Type:** Bool<br />
 **Default:** `true`
 
-For `allOf` inline properties from references
+Controls the behaviour when generating entities from nested `allOf` schemas.
+
+<details>
+<summary>Examples</summary>
+
+With the following schema:
+
+```yaml
+components:
+  schemas:
+    Animal:
+      properties:
+        numberOfLegs:
+          type: integer
+    Dog:
+      allOf:
+      - $ref: '#/components/schemas/Animal'
+      - type: object
+        properties:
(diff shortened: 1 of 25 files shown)


LePips:
Is there additional documentation that can accompany this? Like an example scheme and examples of how they would generate for true and false?

liamnichols:
Good point, I'll add some 👍

PR #141, opened 2022-08-14
Write extension sources into an Extensions subdirectory

liamnichols:
When testing the 0.1 generator out, I decided that I wasn't quite a fan on having extensions left in the top-level of the output directory since it made things look a little messy and disorganised. 

Instead, I'm proposing that we nest them in their own directory named **Extensions** to keep things a little more organised.

Before|After
---|---
Before|After

I also think this aligns a bit better with how other generators work, for example the `openapi-generator` does this (for Android):

Screenshot 2022-08-14 at 22 36 54

WDYT @LePips? When generating with `mergeSources`, they'll continue to be embedded inside **Paths.swift** and **Entities.swift**


LePips:
Think it looks good 👍

PR #142, opened 2022-08-15
Custom mappings between schema formats and Swift Types

LePips:
- Closes https://github.com/CreateAPI/CreateAPI/issues/114

This PR allows custom Swift type overrides for schema datatype formats (not global for this):

```yaml
datatypes:
  string:
    uuid: String
    password: MyPasswordType
  integer:
    int32: Int
    int64: Int
```

Also removes `useFixWidthIntegers` in favor of the above example datatype override.


LePips:
@liamnichols

Because `useFixWidthIntegers` was removed _all_ of the tests need to be updated because it's default was `true`. So, would you like me to either:
1 - update the tests as newly generated with `Int32/Int64` 
    - very large diff, but upholds default behavior. Test to be added would be overriding datatypes
2 - update the test definitions with `int32/int64` overrides
    - smaller diff, but all test datatype overrides. Test to be added would be a plain test for default behavior

liamnichols:
Good point! Let's preserve the original behaviour first (use a default of Int for both of the formats) and then we can sort the default later 👍🏼

Sources/CreateAPI/Generator/Generator+DataTypes.swift
```diff
@@ -0,0 +1,58 @@
...
+                setNaiveDateNeeded()
+                return .builtin("NaiveDate")
+            }
+            return builtInType("String", format: "date", overrides: options.dataTypes.string)
+        case .dateTime:
+            return builtInType("Date", format: "date-time", overrides: options.dataTypes.string)
+        case .other(let format) where format == "uri":
+            return builtInType("URL", format: "uri", overrides: options.dataTypes.string)
+        case .other(let other) where other == "uuid":
+            return builtInType("UUID", format: "uuid", overrides: options.dataTypes.string)
```
liamnichols, line 21: Switch actually lets you write that where clause inline if you want

```suggestion
        case .other("uri"):
            return builtInType("URL", format: "uri", overrides: options.dataTypes.string)
        case .other("uuid"):
            return builtInType("UUID", format: "uuid", overrides: options.dataTypes.string)
```

Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -175,9 +175,6 @@ public struct ConfigOptions: ParsableConfiguration {
     /// Parses dates (e.g. `"2021-09-29"`) using [NaiveDate](https://github.com/CreateAPI/NaiveDate)
     @Option public var useNaiveDate: Bool = true
 
-    /// If enabled, uses `Int64` or `Int32` when specified.
-    @Option public var useFixWidthIntegers: Bool = false
```
liamnichols, line 179: There's a helper in **Issue+Warnings.swift** that instructs the user how to replace this property. 

Please could you update the message to point to the new data types option 👍🏼

liamnichols (commented):
Looking good so far! Yeah let's preserve the original default behaviour for now, then we can change it in a follow up PR 👍🏼

LePips:
To accommodate the smallest diff here for "preserving the default behavior" I force `int32/int64` to be `Int` which only requires one test case to be updated.

LePips:
In my documentation I say that types must conform to `Decodable`. I assume this because I think we can make the decision to conform all generated objects to `Codable` and remove the options `skipRedundantProtocols`, `alwaysIncludeDecodableImplementation`, and `alwaysIncludeEncodableImplementation` unless there is something I am missing.

If there is no other motive for these options, I just think that these are kind of redundant as it's implicitly expected that objects are `Decodable+Encodable`. `Codable` conformance is by default and can be removed manually if desired as always.

Tests/Support/Snapshots/edgecases-int32-int64/Sources/Entities/ArrayTest.swift
```diff
@@ -6,10 +6,10 @@ import NaiveDate
 
 public struct ArrayTest: Codable {
     public var arrayOfString: [String]?
-    public var arrayArrayOfInteger: [[Int64]]?
+    public var arrayArrayOfInteger: [[Int]]?
```
liamnichols, line 9: Can you please update `testEdgecasesEnableIntegerCapacity()` so use `dataTypes` and re-record, that'll undo this diff 🙏

liamnichols (approved):
Briliant work! 🚀

PR #145, opened 2022-08-15
Enable fixed width integers by default and update documentation

liamnichols:
- #142
- #114


liamnichols:
I'm still on the fence as to wether we want this as the default behaviour or not 😅 Before merging I might double check some other generators to see how they handle fixed-width integers by default.

LePips:
I ultimately think whatever is fine. I just have the thought that we should generate as closely to the schema as possible by default and then it's configurable to change that behavior.

liamnichols:
So SwagGen has a similar `fixedWidthIntegers` flag that is `false` by default, and `openapi-generator` uses fixed width integers by default so I'm going to stick with this change and change the default behaviour 👍

PR #150, opened 2022-08-18
Remove Module Namespace from References

LePips:
This simply removes the module namespace from also generating with references. However, this _only_ seems to apply to Paths while it generated since only it seemed to inject the namespace into the context.

I noticed this earlier but didn't think anything about it but noticed it again while moving to our new API. Overall, this is just unnecessary and makes things cleaner without breaking anything.

Unless I missed something. I also think that this can be a bit more desirable if they only generate paths but I can't think of a good reason why.

Before:
```swift
public func get(limit: Int32? = nil) -> Request<[TestAPI.Pet]> { ... }
```

Now:
```swift
public func get(limit: Int32? = nil) -> Request<[Pet]> { ... }
```

Tests are not generated. Overall, I think we can take a good look at the overall generation and clean it up a bit.

### DIFF
--- a/Sources/CreateAPI/Generator/Generator+Schemas.swift
+++ b/Sources/CreateAPI/Generator/Generator+Schemas.swift
@@ -271,7 +271,7 @@ extension Generator {
             }
         }
         name = Template(options.entities.nameTemplate).substitute(name)
-        return .userDefined(name: makeTypeName(name).namespace(context.namespace))
+        return .userDefined(name: makeTypeName(name))
     }
 
     // MARK: - Object


kean:
You need it to avoid name collisions, e.g.

```swift
extension API {
    struct Pet {
        func get(limit: Int32? = nil) -> Request<[Pet]> { ... }
    }
}
```

My initial goal was to make it smart and add namespaces only when necessary, but I encountered some challenges that made it hard to achieve.

LePips:
Makes sense for rest path style. Might be a bit pedantic in wanting it removed, probably just for operations style?

Either way, would probably add another config option?

PR #159, opened 2022-10-04
Improve error messages that are used when `oneOf` object decoding fails

liamnichols:
# Background

- #158 

When we faced the linked bug above, it was hard to figure out exactly what was going on at first. The only message that we saw was the following:

> Failed to intialize `oneOf`

It would be nicer if the message logged could help us to understand what has gone on a little more so I have updated it so that it includes more details about the issue. 

You can see the diff of the snapshots for an idea of how this looks.

### DIFF
--- a/Sources/CreateAPI/Generator/Templates.swift
+++ b/Sources/CreateAPI/Generator/Templates.swift
@@ -301,9 +301,13 @@ final class Templates {
             statements += " "
         }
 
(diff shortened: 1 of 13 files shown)


imjn (approved):
Amazing!!! Thank you very much.
Finally, it's time to adapt DTO and Path to search pages. 
This will save me a lot of time 😄

PR #160, opened 2022-10-08
Update Get dependency to require 2.1.x or later

mgrider:
The only "real" changes here are in `PackageDeclaration.swift` and `Generator+Paths.swift`. The other files are test results that have changed.

I did also test this on a project I'm working on, but I don't feel like that's a great test really, as I'm not sure I'm using all the features of Get or a complex OpenAPI schema.


liamnichols (approved):
Thanks so much for picking this up @mgrider! 

Changes look great to me. In terms of testing, I think that + the CI checks should be plenty as the 2.x version release was only a major release to fix up the API design of `Request` 👍

PR #170, opened 2022-11-23
Enhance support for default values

### LINKED ISSUE #169: Non-optional enum properties with only variant should set a default value
Hello – I'm working on an App Store Connect API client and have found your library very useful for processing the OpenAPI schema. One thing I've noticed that I believe would improve things for my library users would be to simplify these kinds of cases:

```swift
public struct App: Codable, Hashable, Identifiable {
    public var type: `Type`
    ...

    public enum `Type`: String, Codable, CaseIterable {
        case apps
    }
```

Here we have a struct with a non-optional property, referencing an enum with only variant. As a library author and user I would like to minimize the amount of repetition at the callsite, and one way to do that would be to add a default value for this variant in the initializer. Something like this:

```swift
public init(type: `Type` = .apps, ...) {
```

Is that a reasonable enhancement request? I may submit a patch if it isn't too difficult to do.

aaronsky:
Fixes #169.

Breaking changes:
* Booleans are no longer treated as special amongst other properties in the spec. Therefore, if a boolean property is not declared as a `required` property, it may now render as an optional boolean.

Other changes:
* Properties with default values have their default values assigned in initializers and decoders, if their type is a specifically-supported builtin (previously this only worked for booleans).
* Properties with default values that match an enum case will use the enum case as default rather than nil.
* Required properties with enum types that contain only one case now automatically use the only case as a default value.


Sources/CreateAPI/Generator/Generator+Render.swift
```diff
@@ -57,19 +62,36 @@ extension Generator {
                         contents.append(keys)
                     }
                     if decl.protocols.isDecodable, properties.contains(where: { $0.defaultValue != nil }) {
```
aaronsky, line 64: This line evaluates far more often now, and I don't think it's using the right heuristic. Would love to get some feedback.

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -687,10 +687,42 @@ extension Generator {
...
+                        }
+                        // Surround the value in quotes, otherwise the string is rendered improperly.
+                        return "\"\(string)\""
+                    } else if let array = anyValue.value as? [AnyCodable] {
+                        return "\(array.map(describeDefaultValue(_:)))"
+                    } else if let dictionary = anyValue.value as? [String: AnyCodable] {
+                        return "\(dictionary.mapValues(describeDefaultValue(_:)))"
+                    } else {
+                        return nil
+                    }
```
liamnichols, line 714: Nice approach, but I think there are a few scenarios where this doesn't hold up. 

Currently we assume that if there is a `Int`/`Uint`/`Double`/`String`/`Array`/`Dictionary` default value then the property is also that same type, but that is not guaranteed to always be the case.

I see that you accounted for this by checking `formatString`, but additionally a user can override the [`dataTypes`](https://github.com/CreateAPI/CreateAPI/blob/main/Docs/ConfigOptions.md#datatypes) configuration option to switch out types based on formats as they see fit. For example, while a `string` with the `uri` format will default to the `URL` type, the user could disable that by overriding the `dataType` and setting it back to `String`. 

I think a more robust way around this might be to compare the type of the property and then only try to unwrap the default value as the corresponding type if the property can be expressed as a boolean, integer, float or string literals? 

While we can't check for `ExpressibleByXX` conformance directly, we can hardcode a set of types that we know about (`String`, `Double`, `Float`, `Int`, `Int32` etc), so I'm wondering if we can do something like this:

```swift
if type.isExpressibleByStringLiteral, let defaultValue = info?.defaultValue.stringLiteral {
    defaultValue = defaultValue
} else if type.isExpressibleByIntegerLiteral, let defaultValue = ... {
    defaultValue = defaultValue
} else if type.isEnum ... {
    defaultValue = ".\(...)"
}
```

As for Arrays and Dictionaries, I think they might be more complicated than they seem... An array literal seems doable, providing the item schema meets one of the examples above, but admittedly it might not be necessary just yet? As for the dictionary literal, that would only work if the property wasn't a schema object and was instead a literal dictionary literal. I think it's perfectly correct for a property of a json object schema to have a `defaultValue` that is a json object dictionary which would end up generating code that doesn't compile.

---

I haven't dug deep into how you'd go about doing this, but `TypeIdentifier` seems to have some properties already (like `isString`) that will help. Let me know if you have any questions!

Sources/CreateAPI/Generator/Generator+Schemas.swift
```diff
@@ -687,10 +687,42 @@ extension Generator {
...
+                }
+
+                if let value = info?.defaultValue.flatMap(describeDefaultValue(_:)), !value.isEmpty {
+                    let bareValue = value.trimmingCharacters(in: .quotes)
```
liamnichols, line 718: I'm not quite sure if I follow this properly? This is stripping `"` or `'` from the leading/trailing edges of the string but for what reason? Are we trying to escape strings?

liamnichols (commented):
Wow big thank you for putting this Pull Request up so quickly! I've left an inline comment about the approach we are taking for the initialiser as I think there are a couple of other things we need to consider. 

I also want to talk more about using the default value in the decodable implementation because with my understanding of a JSON schema/Open API spec, I don't think that the `default` keyword was defined for this. 

The OpenAPI Docs don't say much about `default`, but they do say it's aligned with [JSON Schema's definition](https://json-schema.org/understanding-json-schema/reference/generic.html), which states the following:

> The `default` keyword specifies a default value. **This value is not used to fill in missing values during the validation process**. Non-validation tools such as documentation generators or form generators **may use this value to give hints to users** about how to use a value. However, `default` is typically used to express that if a value is missing, then the value is semantically the same as if the value was present with the default value. The value of `default` should validate against the schema in which it resides, but that isn’t required.

I can see the `default` value useful as a hit in the generated initialisers, but manipulating the decoded response to fill gaps with it feels like it might be a step too far since nothing suggests that `nullable` and `default` are related to each other (thinking outside of Swift).

That being said, I don't know everything so I could be wrong.. Have you compared how some of the other generators in existence use `default`? I might be swayed if there are other good examples but at the moment I'm leaning towards suggesting that we only apply the default to the initialisers. Let me know what you think, but thanks again for the work here!

Sources/CreateAPI/Generator/Generator+Schemas.swift
aaronsky: Essentially, if the value is a string, we need to wrap it in quotes for rendering. But because defaults correspond to rawValues for enums, in order to maintain that support we need to strip out the quotes so we can check if there is a matching enum. We might be able to do this in `describeDefaultValue` though.

Sources/CreateAPI/Generator/Generator+Schemas.swift
aaronsky: Cool, I'll tinker some more with this.

aaronsky:
> I can see the `default` value useful as a hit in the generated initialisers, but manipulating the decoded response to fill gaps with it feels like it might be a step too far since nothing suggests that `nullable` and `default` are related to each other (thinking outside of Swift).
> 
> That being said, I don't know everything so I could be wrong.. Have you compared how some of the other generators in existence use `default`? I might be swayed if there are other good examples but at the moment I'm leaning towards suggesting that we only apply the default to the initialisers.

I can compare output to some of the other big generators. I generally agree that defaults should not be influencing the decision to generate decoder-inits. I'm also leaning towards minimizing the generation of decoder-inits as much as possible, but that would require us to have a little more insight into whether or not synthesis is possible.

aaronsky:
I've put up a couple new solutions to the feedback you've given me. Let me know what you think of this once you've reviewed, given your time and energy.

Tests/Support/Snapshots/OctoKit/Sources/Entities/Repository.swift
```diff
@@ -169,25 +169,25 @@ public struct Repository: Codable {
...
     /// Whether to allow squash merges for pull requests.
     ///
     /// Example: true
-    public var allowSquashMerge: Bool
+    public var allowSquashMerge: Bool?
     /// Whether to allow Auto-merge to be used on pull requests.
     ///
     /// Example: false
-    public var allowAutoMerge: Bool
+    public var allowAutoMerge: Bool?
```
liamnichols, line 182: Are these changes right? I'm not sure what has caused them but we don't seem t have modified the schema have we?

liamnichols (commented):
@aaronsky sorry it took me so long to get back to this! Just to confirm from your last message, my understanding was that we aren't going to provide a default value when decoding via `init(from:)` and instead would only use the default values in the generated memberwise initialiser?

Tests/Support/Snapshots/OctoKit/Sources/Entities/Repository.swift
aaronsky: My understanding of the OpenAPI spec, limited as it may be, is that it is technically more accurate because these properties are not required in their respective definitions.

aaronsky:
> @aaronsky sorry it took me so long to get back to this! Just to confirm from your last message, my understanding was that we aren't going to provide a default value when decoding via `init(from:)` and instead would only use the default values in the generated memberwise initialiser?

I believe that's what we agreed on.

liamnichols:
> > @aaronsky sorry it took me so long to get back to this! Just to confirm from your last message, my understanding was that we aren't going to provide a default value when decoding via `init(from:)` and instead would only use the default values in the generated memberwise initialiser?
> 
> I believe that's what we agreed on.

Ah ok then, I ask because I still see the nil coalescing to the default value in the snapshots, like this:

https://github.com/CreateAPI/CreateAPI/pull/170/files#diff-e17d0dd37de44a675de95aef390c1a217a05ce891f7bc9902e7b284b75ec5ed0R92

We shouldn't be doing that, right?

aaronsky:
Actually that one is deliberate, because that is the `default` defined in the [github.yml](https://github.com/aaronsky/CreateAPI/blob/89cdceacfb4c0c523945b46e8de8f4ea4045cde9/Tests/Support/Specs/github.yaml) spec

Tests/Support/Snapshots/OctoKit/Sources/Entities/Repository.swift
aaronsky: I'm trying to unwind my changes since it's been a few months since I did most of this work, in order to give a definitive answer on this. It might be possible to split this out, but it's not immediately clear.

liamnichols:
I think that still goes against how we want to interpret `default`? We want to keep it to be provided as a hint for the memberwise initialiser but we don't really want to start making customisations to the decoder behaviour. 

The property is still optional, so if the decoding container doesn't decode the value (either because it was omitted or because it was `null`?) then we should pass that on still rather than falling back to the default. 

With the exception of the separate bug fix, it's important that the only change in generator output is the addition of a default value on the memberwise initialiser rather than behavioural changes to how decoding works. Maybe we could look into seeing if the decoder defaults becomes an opt-in change that we might want to support later on, but I don't think that it should happen by default or in this PR.

Does that sound ok? Thanks for sticking with me on this one as I also try to figure out how we want this to work 😅

aaronsky:
@liamnichols I see what you mean now. I've pushed up a revision to (hopefully) address this.

Sources/CreateAPI/Generator/Templates.swift
```diff
@@ -421,8 +424,7 @@ final class Templates {
         if let metadata = property.metadata {
             output += comments(for: metadata, name: property.name.rawValue, isProperty: true)
         }
-        let isOptional = property.isOptional && property.defaultValue == nil
-        output += "\(access)\(isReadonly ? "let" : "var") \(property.name): \(property.type)\(isOptional ? "?" : "")"
+        output += "\(access)\(isReadonly ? "let" : "var") \(property.name): \(property.type)\(property.isOptional ? "?" : "")"
```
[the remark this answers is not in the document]
liamnichols: Ahh but it does this because the generated decoder implementation was already using the `defaultValue` as a fallback 🙈 

https://github.com/CreateAPI/CreateAPI/blob/da8730b23dab769414229bb386b3dcc2a77d6e42/Sources/CreateAPI/Generator/Templates.swift#L244-L245

Which is what we said that we should stop doing.

I guess this didn't show up much (at all?) before because it was limited to only booleans. Hmmm, now I'm confused as to what we should do because `includeDefaultValues` is `true` by default.

liamnichols (commented):
I looked at the diff a bit closer and now I understand what is going on a bit more, but I am now also more confused 😄 

CreateAPI was already using it's 'default' value in the decoder implementation but I didn't notice this because it only supported boolean defaults from the schema itself.

You can see this here: 

https://github.com/CreateAPI/CreateAPI/blob/da8730b23dab769414229bb386b3dcc2a77d6e42/Tests/Support/Snapshots/edgecases-rename/Sources/Entities/Order.swift#L39

To be honest, this is a really confusing and unexpected behaviour (at least it is to me). If  property is nullable/not required, we're still treating it as non-optional by overriding the Codable implementation to fallback to the default value.

It's kind of neat, but it's kind of also confusing and I am still not sure which direction we should take 😄 

To me, this feels quite non-standard for a code generator to make that kind of decision. Would this in fact encourage poor schema design? Why is a property `nullable` in the first place if there is a `default` value?

Sorry to open a can of worms on this PR given that its existing behaviour, but I guess it's more important as we consider opening it up to more than just booleans.

PR #172, opened 2022-11-28
Support multipart/form-data as opt-in feature

ainame:
### Background 

I want to add multipart/form-data support to CreateAPI. The typical reason people need `multipart/form-data` is the lack of native support for binary data in JSON format. Let’s say you have an endpoint to post not only a comment but also a photo, you would need to send both data in a payload and want to use `multipart/form-data`. Looking at the codebase, currently, CreateAPI seems to postpone supporting the format. When an endpoint supports only `multipart/form-data` in the request body, it becomes Data type. That leaves CreateAPI unusable for such an endpoint. 

https://github.com/CreateAPI/CreateAPI/blob/208445097074b03d429e95756b8819898314982a/Sources/CreateAPI/Generator/Generator%2BPaths.swift#L7
https://github.com/CreateAPI/CreateAPI/blob/d9a7f7d2c78f7d825cfa8ba52f92d973c362204f/Sources/CreateAPI/Generator/Generator%2BSchemas.swift#L6

I can see why it wasn't supported initially because https://github.com/kean/Get doesn't support request body serialization other than `JSONEncoder`. 

### What I changed

I added a new option called `useStructuredMultipartFormDataRequest` to have CreateAPI generate a structured request body, just like normal `application/json` based requests. The default value is `false`. Nothing is affected by existing users here. Previously it was limited to just a Data but now it’s easier to compose data with a documented structure. The idea is that once this was implemented, the responsibility is down to `CreateAPI` users to handle encoding request body. You would need something like `MultipartFormDataEncoder` instead of `JSONEncoder` in their HTTP clients.

This is an example of what a `multipart/form-data` request looks like when the option is enabled.
https://github.com/CreateAPI/CreateAPI/pull/172/commits/31469ade83d93cd2fd5f28aa9ced0ad88fc0eab2#diff-b304b36e06a0c0a28709263be23a00d9ab31e810f924da2a1468e1109e277227

And also, I changed the default type for binary data from `String` to `Data`. Despite the JSON format doesn’t support binary data, `JSONEncoder` and `JSONDecoder` does support `Data` type using Base64 encoding/decoding. `Data` would still work with existing `JSONEncoder` and `JSONDecoder`.

* https://developer.apple.com/documentation/foundation/jsonencoder/2895381-dataencodingstrategy
* https://developer.apple.com/documentation/foundation/jsondecoder/2895108-datadecodingstrategy

This part would be a breaking change. Given the version is still `0.1.1` and theoretically, CreateAPI users have never utilised binary data in their request body just yet (since `multipart/form-data` support isn't in place!), I hope this is an acceptable change🙏  

----

Another thought on using `Data` for binary field. In a `multipart/form-data` request, we normally also need to specify a MIME type. `Data` type alone can do nothing to identify the type. I think we can solve that by leveraging custom `struct` type having both `mimeType` and `data` properties or assuming MIME type from magic bytes, like https://github.com/sendyhalim/Swime.


Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -712,6 +712,10 @@ public struct ConfigOptions: ParsableConfiguration {
         /// Rename anonymous collection elements. By default, use a singularized form of the property name
         @Option public var collectionElements: [String: String] = [:]
     }
+
+    /// If `true`, CreateAPI generates request body structures for "multipart/form-data" format.
+    /// Otherwise such request body fallbacks to `Data` type. The default value is `false`.
```
liamnichols, line 717: So these comments are used to updated documentation too, if you run `make documentation`, it should update **Docs/ConfigOptions.md** with the definition and this content 🙏 

While we are at it, I think we could add some more context to help describe the purpose of this option... How about something like this:

```suggestion
    /// If `true`, CreateAPI generates request body structures for "multipart/form-data" format just like it would for "application/json".
    ///
    /// Otherwise the `body` of the generated `Request` will fallback to `Data`. The default value is `false`.
    ///
    /// When using Get and it's `APIClient`, because Multipart Form Data isn't supported from the `Request` `body` property, it is best to leave this option set to `false`. 
    /// If however you have implemented your API Client, and you prefer to use structured `Codable` types to encode a Multipart Form Data request body, setting this value to `true` can be more convenient. 
    /// You might also need to use the [`dataTypes`](#datatypes) option to customise the type used to represent `binary` data.
```

Sources/CreateAPI/Generator/Generator+DataTypes.swift
```diff
@@ -7,6 +7,8 @@ extension Generator {
         switch format {
         case .byte:
             return builtInType("Data", format: "byte", overrides: options.dataTypes.string)
+        case .binary:
+            return builtInType("Data", format: "binary", overrides: options.dataTypes.string)
```
liamnichols, line 11: I think this is fine. It's a breaking change but it's minor and there is a way to revert to previous behaviour. I'll add a note in the release notes:

> `string` properties in a schema with the `format` set to `binary` will now be generated as `Data` types instead of `String`. To revert back to the previous behaviour, use the [`dataTypes`](https://github.com/CreateAPI/CreateAPI/blob/main/Docs/ConfigOptions.md#datatypes) configuration option:
>
> ```yml
> dataTypes:
>   string:
>     binary: String
> ```

liamnichols (commented):
This is great! Thanks for the contribution 🚀 I just added some notes about naming of the property and docs, but I am looking forward to shipping this as part of 0.2.0 🙌

Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -712,6 +712,10 @@ public struct ConfigOptions: ParsableConfiguration {
...
+    @Option public var useStructuredMultipartFormDataRequest: Bool = false
```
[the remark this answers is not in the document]
ainame: > Alternatively, I wonder if its clearer if we flip the purpose (and default it to true) by setting it to useDataForMultipartFormDataRequestBodies? While I understand what 'structured' means in this sense, unless I knew that using Data was being referred to as 'unstructured', I might not have properly figured out what the original meaning was.

This makes sense. The only concern from me in making the new behaviour opt-out was how much we should preserve the backward compatibility. If it's not the case, I favour flipping the option!

Sources/CreateOptions/ConfigOptions.swift
ainame: Adopted this as well as flipping true and false for new option name

Docs/ConfigOptions.md
```diff
@@ -103,6 +103,7 @@ Below you can find the complete documentation for all available options.
   - [entities](#renameentities)
   - [operations](#renameoperations)
   - [collectionElements](#renamecollectionelements)
+- [useDataForMultipartFormDataRequestBody](#usedataformultipartformdatarequestbody)
```
ainame, line 106: Maybe the option should have been a part of `paths` 🤔

ainame:
I noticed that the option only applies to `Paths`, so I moved it to under `paths` section.

https://github.com/CreateAPI/CreateAPI/pull/172/commits/23d4a39b5b17808a7873dc46c4bcf38b2716b044

liamnichols (approved):
Thanks @ainame! These changes look good to me 🚀 

Is there anything else that needs doing to support this or is it ready to merge?

ainame:
No. The blocker was only me figuring out how to encode an `Encodable` `struct` to `multipart/form-data` format. I was able to do that with this change. I'm confident it's working now.

PR #175, opened 2022-12-19
Sort Properties After Creation

### LINKED ISSUE #174: `useSwiftyPropertyNames` and `sortPropertiesAlphabetically` ordering
Using `useSwiftyPropertyNames: true` and `entities.sortPropertiesAlphabetically: true`, the `is-` conversion is happening after the alphabetical sorting.

Example schema:

```yaml
// ...
components:
  schemas:
    Pet:
      title: A pet title
      description: A pet description
      type: object
      required:
        - id
      properties:
        id:
          type: string
        likesMeat:
          type: boolean
        allergic:
          type: boolean

```

Generates the following:

```swift
public struct Pet: Codable {
    public var isAllergic: Bool?
    public var id: String
    public var isLikesMeat: Bool?

    // ...
}
```

It would be much more preferable if properties were properly sorted:

```swift
public struct Pet: Codable {
    public var id: String
    public var isAllergic: Bool?
    public var isLikesMeat: Bool?

    // ...
}
```

Additionally, it would probably be beneficial for an exclude list for `useSwiftyPropertyNames` as some property conversions may not be grammatically preferable, like in the above example.

LePips:
- Closes https://github.com/CreateAPI/CreateAPI/issues/174

Sorts created properties, which go through their own independent renamings if applicable, instead of the raw keys.


LePips:
Interesting, the tests' type checker couldn't do that inference, I think due to `handle(error:_)`.

LePips:
@liamnichols I apologize for the ping, could this and #177 receive a review as I would like to go ahead with some work that would use these features. Thank you!

liamnichols (commented):
Hi @LePips, really sorry about the delay in getting though these notifications! Thanks for taking the time to work on this, I really appreciate it! 

The changes looks good to me assuming that it is doing what you say it does, but could you maybe replicate it in the tests somehow so that we can be sure not to break the behaviour in the future by mistake? 🙏

PR #177, opened 2022-12-19
Entity Property Type Override

### LINKED ISSUE #176: Entity Individual Property Swift Type Override
While property schema `datatype`'s can be overridden, I need to be able to override a single entity property type.

---

### Context

The schema I am using has a string property that is represented by a string enum on the backend however isn't specified in the schema as an enum due to the complexity/time of rewriting the backend using that enum (along with a few others).

So, I have my own Swift enum specified and would like to use that instead.

Currently the following would be generated:
```swift
struct MyItem {
	let itemType: String
}
```

What should come from generation:
```swift
struct MyItem {
	let itemType: MyItemType
}
```

---

### Implementation

Introduce new config option `entities.propertyTypeOverrides` which represents a mapping of scheme properties to Swift types,

```yaml
entities:
  propertyTypeOverrides:
    MyItem.itemType: MyItemType
```

This option would take precedence over the datatype override and would follow the same rule:

> It is your responsibility to ensure that the replacement type conforms to Codable and can properly decode and encode to the original primitive type.

LePips:
- Closes https://github.com/CreateAPI/CreateAPI/issues/176

Implements individual property type override. This allows both general property naming and specific property type, mirroring the usage of `rename.properties`. I consider this an advanced feature, since this allows such fine control over single properties.

### Overriding Optional

Since this is an advanced feature and by definition allows one to override the _actual property type_, it could be thought that his encompasses the _entire_ type which includes whether it is an `Optional`.

For example, assume that `Pet.tag` would originally map to `String?`, but with the below:

```yaml
entities:
  propertyTypeOverrides:
    Pet.tag: UUID
```

this would result in `Pet.tag` mapping to `UUID`, therefore making this required to the client. This can also happen the other way, making a non-optional property generate as an optional instead. This would also of course be a responsibility to the developer that the types can properly decode/encode.

I don't do this here (`Optional` is stable through the type override), however I think this should be a consideration.


mgrider:
FWIW, this looks good to me. 🤷‍♂️

Issue #171, opened 2022-11-23
Support Hashable


sjmueller:
We have the requirement that all of our entities and enums implement `Hashable`. For the entities this is mostly achievable by setting our `config.yaml`:

``` yaml
entities:
  protocols: [Hashable, Codable]
```

However, two problems arise with this configuration:
1. There's no way to set `Hashable` conformance to enums, which breaks all Entities with an enum property at compile time: `Entity does not conform to Hashable`.  This can be worked around by manually extending all enums:
  ``` swift
  extension UserStatus: Hashable {}
  ```
2. If an entity has a map property, then create-api will generate an `enum AnyJSON` which also doesn't conform to `Hashable`. However, using the same extension workaround yields compile time error:
```
Extension outside of file declaring enum 'AnyJSON' prevents automatic synthesis of 'hash(into:)' for protocol 'Hashable'
```

## Potential Solution
Add a configuration option that allows conformance for enums, in the same way that is afforded for entities:
``` yaml
enums:
  protocols: [Hashable, Codable]
```

There's also the argument that `Hashable` should be universally used over `Equatable` by default in this library, since all `Hashable` constructs are also `Equatable`.

aaronsky:
Likewise, given recent developments in Swift 5.8, I would like to see additional conformance to Sendable in enums and other places. Otherwise, it is impossible to add proper Sendable conformance through the `entities.protocols` configuration.

liamnichols:
Thanks for sharing this request! I agree that this would be a great addition to allowing us to customise enum generation 🙏 A similar configuration style was also proposed here: https://github.com/CreateAPI/CreateAPI/issues/130#issuecomment-1214136925 so I think the `enums` option is a right approach to cover use cases like the one that @aaronsky raised with `Sendable` too. 

It would be interesting to dig into `Hashable` by default, but there are some considerations to be had when using the generator to produce class types. We'd have to dig a bit further into that one, but in the meantime I think adding the `enums.protocols` support would be a good start 👍

Issue #184, opened 2023-02-24
integer backed enums are stored as integer properties as opposed to enums


mgrider:
The issue is that an OpenAPI definition containing something like:

```yaml
        enum_integer:
          type: integer
          format: int32
          enum:
            - 1
            - 2
```

...will result in code that looks like:

```Swift
    public var enumInteger: Int?
```

That example is from the unit tests: https://github.com/CreateAPI/CreateAPI/blob/0069e8b87e7ea1dd1dc508f271de7d3ddaa7eb0e/Tests/Support/Specs/edgecases.yaml#L1097-L1102

I would prefer/expect to see something more like:

```Swift
    public enum EnumInteger: Int, Codable {
        case _1 = 1
        case _2 = 2
    }
```

I'm assuming this was a choice, probably? But maybe there could be a flag for it or something?

mgrider:
No doubt the negative number complicates things. 😂

PR #185, opened 2023-03-10
Added support for enum.protocols feature

aaronsky:
Addresses #171.

Creates a new `enum.protocols` option that is used to drive protocol conformance for enums of strings, which are the only currently supported enum variety. Currently does not do any of the redundant protocol stripping that `entities.protocols` does.


Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -568,6 +568,18 @@ public struct ConfigOptions: ParsableConfiguration {
         @Option public var filenameTemplate: String = "%0.swift"
     }
 
+    @Option public var enums: Enums
+
+    // sourcery: document,
+    /// Options specifically related to generating enums of strings.
```
liamnichols, line 574: Just a thought, but #184 might mean that we generated enums from integer properties too so maybe worth rewording this:

```suggestion
    /// Options specifically related to enums generated from schemas.
```

Sources/CreateOptions/ConfigOptions.swift
```diff
@@ -568,6 +568,18 @@ public struct ConfigOptions: ParsableConfiguration {
...
+    ///
+    /// Note: these options do not apply to other enums, such as `CodingKey` conformances or
```
liamnichols, line 576: Just a suggestion to get the nicer formatting in GitHub

```suggestion
    /// > **Note**: these options do not apply to other enums, such as `CodingKey` conformances or
```

It would look like this:

> **Note**: these options do not apply to other enums, such as `CodingKey` conformances or `ResponseHeaders`.

liamnichols (commented):
Wow great work! Thanks for this! 

> Currently does not do any of the redundant protocol stripping that `entities.protocols` does.

I think that is fine. This is a good start 🙏 

--- 

Looks like there are some issues with the CI checks, but I'd love to get this merged once that is addressed 🙏

aaronsky:
@liamnichols I believe this both addresses the test failures and your feedback. The PR suddenly got huge because I needed to regenerate snapshots, since protocols are sorted by name now.

aaronsky:
@liamnichols any update on this?

damien-nd:
Any update on this PR? I would like to add Sendable on enum like we can do on entities.protocols 🙏

nbapps:
Any update on this ?

mgrider:
I'll add my voice to the chorus here. It would be great to get this merged.

mgrider:
~~...although looking closer, I can see some reasons we wouldn't want this version of this feature. For one thing, why is `enums` expected to be a separate part of the config from `entities`? It should be parallel (at the same level as) `protocols` underneath `entities`, IMO.~~ (Ignore me here, I was just confused.)

Also, I'm not sure why this PR had to regenerate all those test results. That's probably something that should be figured out before this could get merged.

aaronsky:
As I stated [here](https://github.com/CreateAPI/CreateAPI/pull/185#issuecomment-1474436956), snapshots needed regeneration because of [this change](https://github.com/CreateAPI/CreateAPI/pull/185/files#diff-eaaa0a74007e890b2d33806e4392b33e8818615b98d81a4d087142b8963df0c1R99)

PR #190, opened 2023-05-24
ConfigOptions: add new option rename.sanitizeRenames

AndrewSB:
incomplete, just prototyping for my own usage right now.

would love an initial pass from someone to let me know if my approach is overall in the right direction


LePips:
Can you give examples of expected output given a sample schema?

AndrewSB:
in source? or just here?

an example is an entity rename of `Measurement` to `Measurement<UnitDuration>`. that's the test i added'

LePips:
I would recommend adding a simple example to your description to know what the problem is.

This will most likely require a test case in `GenerateOptionsTest` with the sanitisation enabled with a rename.

Issue #173, opened 2022-12-09
Consider GenericJSON over AnyJSON?


sjmueller:
In have some postgres json fields in our tables with null values:

``` json
{
    "theme": null,
    "url": "https://blah.com/etc.png"
}
```

This crashes CreateAPI decode because `AnyJSON` isn't able to handle nulls. Have you considered Using [GenericJSON](https://github.com/iwill/generic-json-swift) instead? It has a clever approach that allows for more flexibility, including nulls.

Here's how we've tried to work around the problem:
``` swift
typealias AnyJSON = GenericJSON.JSON // doesn't work, just creates ambiguity
```

``` yaml
# create-api.yaml
entities:
  exclude: [AnyJSON] # Doesn't work, ignored
```

Finally what worked for us:
- create a stub `Json` schema on the server api spec, and use everywhere we have a json field
- `exclude: [Json]` in the config
- `typealias Json = GenericJSON.JSON` to shoehorn in the desired implementation

Is there a better way?

liamnichols:
Thanks for raising this! That's an interesting point. `AnyJSON` is a special type which is why you aren't having any luck with `entities.exclude` (its for excluding your own schema entities).

I guess I am curious about the root issue of "This crashes CreateAPI decode because AnyJSON isn't able to handle nulls."... Do you have a particular example of this? 

I'm just trying to understand why your generated code isn't generated as `AnyJSON?` which should handle the null type.. Is it because it is nested as `[String: AnyJSON]`? I was wondering if:

a. It should actually be `AnyJSON?` because `null` is valid JSON after all
b. `AnyJSON` should have a `null` case

I'm not sure if either of those approaches will have other complications though (i.e `Optional<AnyJSON>.none` vs `Optional<AnyJSON>.some(.null)` and so on 🤔

LePips:
Could the issue be that the fields aren't marked as `nullable` in the schema?

LePips:
It does look like `AnyJSON` does not support `nil/null` decoding, just by looking at the `AnyJSON` implementation and by a basic test:

```swift
let encoder = JSONEncoder()

let json: [String: Int?] = ["test": nil, "asdf": 123]

let data = try encoder.encode(json)

let decoder = JSONDecoder()

let result = try decoder.decode(AnyJSON.self, from: data)

print(result)
```

LePips:
[...]

PR #200, opened 2023-07-27
AnyJSON null support

grishamsc:
Add support for `null` values in free-form JSON value.
```
{
    "decoded_body": {
        "query_id": 1690409533,
        "amount": "1000000000",
        "destination": "0:25603f6d7d3a1c6f981f02237c917150a6f2af971e83dd9e19605fa57a5d5b00",
        "response_destination": "0:25603f6d7d3a1c6f981f02237c917150a6f2af971e83dd9e19605fa57a5d5b00",
        "custom_payload": null,
        "forward_ton_amount": "1",
        "forward_payload": {
        "is_right": false,
        "value": "b5ee9c7201010101000300000140"
        }
    }
}
```

### DIFF
--- a/Sources/CreateAPI/Generator/Templates.swift
+++ b/Sources/CreateAPI/Generator/Templates.swift
@@ -638,14 +638,16 @@ final class Templates {
     var anyJSON: String {
         """
         \(access)enum AnyJSON: Equatable, Codable {
+            case null
             case string(String)
             case number(Double)
             case object([String: AnyJSON])
             case array([AnyJSON])
             case bool(Bool)
 
-            var value: Any {
+            var value: Any? {
                 switch self {
+                case .null: return nil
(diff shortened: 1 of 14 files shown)


AndrewSB:
did something change recently? i'm just getting this error now. i don't think it was a problem a couple weeks ago

LePips:
Related: #173

AndrewSB:
+1 on this, i was trying to encode a new type and was getting a crash that i couldn't figure out for a while

i'm going to use the `exclude: [Json]` and import GenericJSON for myself, like @sjmueller described in #173, but doing any one of the 3 following seem like a good idea: 

1. merging this to support null
2. having a special error message that fires, linking to this issue or #173, instead of crashing inside JSONEncoder with an EXC_BREAKPOINT
3. providing an official override for Json

